PR #16, opened 2024-01-18
Initial repository structure

misohu:
Prepare repository according to patterns used in [Chisme repo](https://github.com/canonical/charmed-kubeflow-chisme/blob/main/tox.ini). Design choices: 
- we will use nested modules structure under src (same as Chisme)
- we will use tox environments similar to other repos
- we will use pip compile to handle tox dependencies (same as other repos)

Note: checks are meant to fail as there is no real code yet.


DnPlas (approved):
Thanks @misohu, nothing much to say if the PR is populating the repo. We can merge it in red because there is nothing to test yet.

Issue #25, opened 2024-02-12
Create a snapcraft.yaml file for the DSS


DnPlas:
### Why it needs to get done

A `snapcraft.yaml` file must be created defining all what is needed for the DSS to be installed, started, removed, etc.

### What needs to get done

To write a `snapcraft.yaml`.  The actual content of the snap depends on the exploration task #23, this field will be updated once we have worked on it.

Reference:
* https://snapcraft.io/first-snap/python/linux/package
* https://snapcraft.io/build

### When is the task considered done

TODO: depends on #23 to clearly define what are the contents of the `snapcraft.yaml` file and deliverables of this task.

DnPlas:
Following the docs for [creating a snap](https://snapcraft.io/docs/creating-a-snap), here are some details about the implementation.

### [Snapcraft checklist](https://snapcraft.io/docs/snapcraft-checklist)

1. Language: python3
2. Toolkits and desktop support: not needed
3. System integration:
* Access to the KUBECONFIG file, usually placed in `~/.kube/config`. The only thing that could be a problem with this is that snap interfaces usually access non-hidden files. To workaround this we may need to place a non-hidden file in the user's home directory. The [home interface](https://snapcraft.io/docs/home-interface) will help out with this.

DnPlas:
List of nice to have:

* [tab completion](https://snapcraft.io/docs/tab-completion)

DnPlas:
Design choices:

1. We need to create our own plugin for accessing hidden files in `$HOME/.dss/config` to be able to get the KUBECONFIG from there. Users will have to save it there always.

```
plugs:
  dot-dss-config:
    interface: personal-files
    read:
      - $HOME/.dss/config
 ```

This requires uses to connect manually with `sudo snap connect dss:dot-dss-config`. This is also something we'll have to document.

2. The snap version is picked up automatically from Github tags. This will help us when automating the release process.
3. We'll use both the `home` and the `network` interfaces. The former allows us to access files in the home directory and the latter allows `dss` to access the microk8s API.

PR #29, opened 2024-02-15
Implement initialize

### LINKED ISSUE #27: Implement DSS initialize command
### Why it needs to get done

Running `dss initialize` will create a Kubernetes `dss` namespace for all user notebooks and it will deploy MLflow . All Kubernetes components connected with DSS will be deployed under the `dss` Kubernetes namespace. The initialize command accepts a `--kubeconfig` parameter which will be used for accessing the underlying Kubernetes cluster. If the `kubeconfig` is not provided, the initialize command will search for it in the `KUBECONFIG` environment variable.

### MLflow
To deploy MLflow properly in local mode using the `initialize` command, DSS will generate the following Kubernetes objects:
- MLflow-data Persistent Volume Claim (PVC) backed by the hostPath storage class.
- MLflow deployment responsible for the MLflow server. This deployment also mounts the aforementioned PVC on /mldata to ensure data persistence in the event of a Microk8s restart. For more information on how to run the MLflow server in local mode refer to [this guide](https://mlflow.org/docs/latest/getting-started/intro-quickstart/index.html).
- MLflow service of type ClusterIP, exposing the MLflow server.

### What needs to get done

1. Implement initialize command which deploys aforementioned components.

1. Command is implemented
2. Tests are passing

misohu:
This PR implements https://github.com/canonical/data-science-stack/issues/27. To test locally you can check integration tests or:

1. Install microk8s with hostpath-storage addon 
2. Clone repo and install the library with `pip install .`
3. Run dss initialize 
```
dss initialize --kubeconfig="$(microk8s config)"
```
4. Dss namespace should be crated and `mlflow-deployment` should be present inside it.
5. Search for `mlflow-svc` in `dss` namespace and use the ip address to access mlflow ui in browser.

**Note:** there needs to be echo as its multi line output.

Closes #27


src/dss/manifests.yaml
```diff
@@ -0,0 +1,61 @@
...
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
+
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
```
kimwnasptd, line 23: Putting this comment here, but can be the case for the rest of the YAML objects we apply:

Let's add common labels across all objects that are part of a "component". In cases of `Deployments` let's add those in both `.metadata` and in `.spec.selector.matchLabels`. This is also what we do in Kubeflow
https://github.com/kubeflow/manifests/blob/6cb9ff2fa2b25755eca5ce41c03d125aa8c8653b/contrib/kserve/kserve/kserve_kubeflow.yaml#L20523

I propose the following labels for the MLFlow components (Deployment, SVC)
- `app: dss-mlflow`
- `app.kubernetes.io/name: dss-mlflow`

tests/integration/test_dss.py
```diff
@@ -0,0 +1,42 @@
...
+    kubectl_result = subprocess.run(
+        ["kubectl", "get", "namespace", "dss"], capture_output=True, text=True
+    )
+
+    # Check if the namespace exists
+    assert "dss" in kubectl_result.stdout
+
+    # Check if the mlflow-deployment deployment is active in the dss namespace
+    kubectl_result = subprocess.run(
+        ["kubectl", "get", "deployment", "mlflow-deployment", "-n", "dss"],
```
kimwnasptd, line 33: Reminder to change the name here as well (or better use `-l <label>` in kubectl.

src/dss/main.py
```diff
@@ -0,0 +1,50 @@
...
+)
+def initialize_command(kubeconfig: str) -> None:
+    """
+    Initialize the Kubernetes cluster and deploy charms.
+    """
+    logger.info("Executing initialize command")
+
+    # If kubeconfig is provided, set it as an environment variable
+    if kubeconfig:
+        os.environ[KUBECONFIG_ENV_VAR] = kubeconfig
```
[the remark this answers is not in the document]
ca-scribner: Or if the user wants to execute something from a second terminal

src/dss/manifests.yaml
ca-scribner: @misohu does have some already, although slightly different.  The `KubernetesResourceHandler` [applies labels](https://github.com/canonical/data-science-stack/pull/29/files#diff-601d3d8b56c9ee039324597dae9fbef6a90af671d96f5063b8f64ae1a840c7d7R79) to everything it deploys, although their design intent was more for letting the KRH easily clean everything up later.  I think it works for this purpose here too though.

I wonder if we should split these resources up into more than one yaml though.  I don't think the `dss` namespace should have a `mlflow` label.

src/dss/initialize.py
```diff
@@ -0,0 +1,104 @@
...
+        field_manager="dss",
+        labels={"app": "dss", "dss-component": "mlflow"},
+        template_files=[manifests_file],
+        context={},
+        resource_types={Deployment, Service, PersistentVolumeClaim, Namespace},
+    )
+
+    # Load Kubernetes generic resources
+    load_in_cluster_generic_resources(client)
+
```
ca-scribner, line 87: afaict this is not needed.  I'll test this later

src/dss/manifests.yaml
ca-scribner: Actually the labels from the KubernetesResourceHandler wont quite match what you're asking, as they will not be in the Deployment's pod template

ca-scribner: @kimwnasptd any reason why we need both `app` and `app.kubernetes.io/name`?  This feels redundant

tests/integration/test_dss.py
```diff
@@ -0,0 +1,42 @@
+import subprocess
+
+
+def test_initialize_creates_dss() -> None:
```
ca-scribner, line 4: these tests do not clean up after themselves, meaning this could happen:

* run tests
  * tests pass (as expected)
* tests create namespace, deployment, etc successfully
* change the code in a way that breaks the tests (for example, edit manifest.yaml and delete the deployment)
* run tests again
  * tests pass (not as expected)

src/dss/main.py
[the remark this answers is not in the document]
ca-scribner: I believe we need to rethink the environment variable as storage.  This sounds like it only works for a subset of where we need it to. 

Can we just cache the file somewhere?  As a standalone python package I'm not sure where we'd put the file, but as a snap (which this will be deployed as) I think we have somewhere we can store files right?  Maybe we just write the config that is provided to that location?

src/dss/initialize.py
ca-scribner: Removed

src/dss/manifests.yaml
```diff
@@ -0,0 +1,61 @@
...
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: dss
+
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: mlflow-pvc
```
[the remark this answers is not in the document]
ca-scribner: Updated.  Look good now?

src/dss/manifests.yaml
ca-scribner: Added some but slightly different than you requested @kimwnasptd.  wdyt?

tests/integration/test_dss.py
ca-scribner: Updated

tests/integration/test_dss.py
ca-scribner: Added a fixture that does this, albeit crudely.  Maybe there's a better way we could do in future, but this is a start

src/dss/manifests.yaml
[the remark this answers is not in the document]
kimwnasptd: Looks good!

src/dss/main.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+@main.command(name="initialize")
+@click.option(
+    "--kubeconfig",
+    help="Specify the kubeconfig content. Defaults to the value of file at path on the KUBECONFIG environment variable.",  # noqa E501
+)
+def initialize_command(kubeconfig: str) -> None:
+    """
+    Initialize the Kubernetes cluster and deploy charms.
```
kimwnasptd, line 27: Can we also rename this? The cli does not deploy any charms

src/dss/main.py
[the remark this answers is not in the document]
kimwnasptd: Agree with writing it to a file. This is also the only way other commands in the future can use this context to interact with the cluster.

I'd suggest the following:
1. Let's use an env var for the path we store this file (allows for configuration for dev env vs snap env)
    1. When the code is part of a snap maybe this will be `$HOME/snap/dss/common` https://ubuntu.com/robotics/docs/snap-data-and-file-storage
    2. For dev purposes, it could be a place here in the repo?
2. Later on once we'll look into the snapcraft structure we could look into a way to change this env var when calling the CLI https://github.com/canonical/data-science-stack/issues/25

@ca-scribner let me know what you think. And if we then decide to go with what proposed in 2 let's then add a comment in that issue to not forget we'll need this

src/dss/initialize.py
```diff
@@ -0,0 +1,103 @@
...
+from dss.logger import setup_logger
+
+# Set up logger
+logger = setup_logger("logs/dss.log")
+
+
+DSS_CLI_MANAGER_LABELS = {"app.kubernetes.io/managed-by": "dss-cli"}
+
+
+def wait_for_deployment_ready(
```
kimwnasptd, line 19: Should we move this to a different file with common code? I would expect other future commands to need to use this (i.e. when creating a notebook/service)

src/dss/main.py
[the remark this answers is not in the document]
ca-scribner: Yeah I was thinking something similar.  Its a bit tricky though since our source is `microk8s config` rather than an existing file.  `kubectl` does something a bit different as it accepts a path to a config file (and doesn't cache that path).  I wonder if we need two separate things:
* an environment variable that defines where we look for a kubeconfig, which by default could be the place we cache a previously provided kubeconfig
* a way to give the dss tool a kubeconfig file to cache (either as a flag to initialize, or as a separate command)

I'll play with this a bit then we can discuss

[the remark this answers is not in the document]
ca-scribner: Refactored to get something using file paths rather than content.  wdyt?  We can rework the defaults when we know more about how snaps can handle this

src/dss/main.py
ca-scribner: done

src/dss/initialize.py
ca-scribner: done

kimwnasptd:
@ca-scribner the PR looks good as a first iteration!

I'd suggest that we keep the following in mind for next steps:
1. For now the `dss initialize` command will only error because of a timeout, but the message will not be helpful on debugging
    1. Let's open bugs anytime we will hit the timeout, and document why it failed. So that we'll gradually improve the error handling
    2. Let's also ask the UX team what the expectation is when this command will fail. Should we urge users to then run `dss status` or `dss logs` in the message so that they know how to further go and fix the issue?
2. ACK for now to just use paths for the kubeconfig, but lets have an issue to document this current limitation (and how it potentially affects the other CLI commands that need a kubeconfig) so that we keep track of it and while working on the snap (and potentially link the issues together)

cc @DnPlas

kimwnasptd (approved):

src/dss/main.py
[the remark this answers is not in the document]
kimwnasptd: Sounds good, added some suggestions for next steps
https://github.com/canonical/data-science-stack/pull/29#issuecomment-1968863944

ca-scribner:
agreed @kimwnasptd, added #37 to capture the `kubeconfig` task

PR #40, opened 2024-03-04
Refactor `dss initialize` to make code easier to reuse

ca-scribner:
* Moves constants to the shared `config.py` file so their definitions are shared
* moves the `manifests.yaml` for dss's namespace + mlflow deployment to a `manifest_templates` directory and splits the namespace out from the mlflow resources
* adds templating to the existing manifests to remove hardcoded values
* removes a leftover, unused fixture

### DIFF
--- a/tests/integration/test_dss.py
+++ b/tests/integration/test_dss.py
@@ -5,7 +5,7 @@ from charmed_kubeflow_chisme.kubernetes import KubernetesResourceHandler
 from lightkube.resources.apps_v1 import Deployment
 from lightkube.resources.core_v1 import Namespace, PersistentVolumeClaim, Service
 
-from dss.utils import DSS_CLI_MANAGER_LABELS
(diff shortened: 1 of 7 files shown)


ca-scribner:
fyi, this PR was reviewed in parallel to #41.  Some minor comments in #41 were actually incorporated here instead (and #41 rebased onto here to receive the changes).

tests/integration/test_dss.py
```diff
@@ -5,7 +5,7 @@
 from lightkube.resources.apps_v1 import Deployment
 from lightkube.resources.core_v1 import Namespace, PersistentVolumeClaim, Service
 
-from dss.initialize import DSS_CLI_MANAGER_LABELS
+from dss.utils import DSS_CLI_MANAGER_LABELS
```
DnPlas, line 8:
```suggestion
from dss.config import DSS_CLI_MANAGER_LABELS
```

?

Issue #31, opened 2024-02-26
Implement DSS `create-notebook` command


DnPlas:
### Why it needs to get done

The create-notebook command allows users to create a Jupyter notebook within the DSS environment. Users specify the notebook's name and an image for the notebook. Behind the scenes, the DSS Python library creates a notebook object from a Kubernetes manifest template, using the provided name and image flags. The command waits until the notebook pod is ready and then outputs a link to access the notebook UI. While waiting for the pod, DSS will output the status of the Pod (Creating, Waiting …). This command will fail with an error message if the Pod won’t get created or if it ends with an error.

The command looks like:

```
dss create-notebook --name user-notebook --image kubeflownotebookswg/jupyter-scipy:v1.8.0
```

Whenever a notebook is created dss deploys these Kubernetes objects: 
* Notebook deployment. This notebook will mount the `user-data` PVC which is shared across all notebooks. Additionally, the notebook requires an environment variable `MLFLOW_TRACKING_URI` set with a URL pointing to the MLflow server to be able to connect.
* Notebook ClusterIP service, providing access to the notebook. 

```
Output:
Access the notebook at http://10.152.183.223/notebook/user-namespace/user-notebook/
```

This command has also a -h or –help  flag which prints out the help description of the command with a list of recommended images for notebooks. For this list we will use the default list from[ jupyter-ui](https://github.com/canonical/notebook-operators/blob/main/charms/jupyter-ui/config.yaml#L27). This list can be subject to a change based on customer needs.
Valid image: is the image which is accessible by DSS and can create a valid jupyter notebook (you can read more about creating valid images in the Documentation section). If the image should support GPU enhanced workloads the image must contain CUDA and cuDNN drivers (or tools needed for Intel GPU based workloads). The image may not contain the MLflow client python library.

### What needs to get done

1. Implement `create-notebook` command which deploys aforementioned components.
2. Implement integration test with Microk8s

### When is the task considered done

1. Command is implemented
2. Tests are passing

ca-scribner:
@DnPlas can you clarify what this command should be creating?  The issue mentions "creates a notebook object from a Kubernetes manifest template" but I think what we planned was to create a `deployment` that deploys a jupyter notebook (creating a `notebook` object means dss also deploys the kubeflow notebooks operator, which I think is out of scope).  Am I misunderstanding this one?

ca-scribner:
[...]

PR #41, opened 2024-03-04
Add creation of notebooks pvc to `dss initialize`

ca-scribner:
This adds to `dss initialize` the deployment of the single PVC that all notebooks share.  This was missed in #29.

Unblocks #31 
Review after #40


src/dss/manifest_templates/mlflow_deployment.yaml.j2
```diff
@@ -1,19 +1,12 @@
...
-  labels:
-    app.kubernetes.io/part-of: dss
 
 ---
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
-  name: mlflow
-  namespace: dss
+  name: {{ name }}
```
DnPlas, line 6: If in the future we add some other manifest files that need to be render with a `name` and `namespace` different than these, things will collide. I think it would be nice to be clear about this context variables, maybe prefix them with `mlflow_`?

src/dss/initialize.py
```diff
@@ -26,14 +31,22 @@ def initialize(lightkube_client: Client) -> None:
...
     # Path to the manifests YAML file
-    manifests_file = os.path.join(os.path.dirname(__file__), "manifests.yaml")
+    manifests_files = [
+        Path(Path(__file__).parent, MANIFEST_TEMPLATES_LOCATION, "dss_core.yaml.j2"),
+        Path(Path(__file__).parent, MANIFEST_TEMPLATES_LOCATION, "mlflow_deployment.yaml.j2"),
+    ]
+
+    config = {
+        "name": MLFLOW_DEPLOYMENT_NAME,
+        "namespace": DSS_NAMESPACE,
```
DnPlas, line 41: Will everything be deployed in the same namespace?

src/dss/utils.py
```diff
@@ -14,6 +14,9 @@
 KUBECONFIG_ENV_VAR = "DSS_KUBECONFIG"
 KUBECONFIG_DEFAULT = "./kubeconfig"
 
+# Labels applied to any Kubernetes objects managed by the DSS CLI
+DSS_CLI_MANAGER_LABELS = {"app.kubernetes.io/managed-by": "dss-cli"}
```
DnPlas, line 18: Could you import this from `config.py`?

DnPlas (commented):
Thanks @ca-scribner, tiny comments.

src/dss/initialize.py
ca-scribner: yeah, I think the spec says everything in dss is in a single namespace.  Unless I misunderstood

src/dss/manifest_templates/mlflow_deployment.yaml.j2
ca-scribner: yeah this is a good point.  If they're rendered by different KRH's then it doesn't really matter, but part of the KRH conveneince is bundling them all together.  I can add prefixes

src/dss/manifest_templates/dss_core.yaml.j2
```diff
@@ -5,3 +5,17 @@ metadata:
...
+kind: PersistentVolumeClaim
+metadata:
+  name: {{ notebook_pvc_name }}
+  namespace: {{ namespace }}
+spec:
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
```
kimwnasptd, line 19: This worries me a bit. For now because of microK8s architecture it won't matter because the PVC will grow outside this limit either way.

I'd suggest that we move this to 10Gi early on, so even when we have to make the update in the future it won't bite us that much that users will only be getting 1Gi by default if the underlying logic of microK8s changes

ca-scribner: Happy to change it, but I don't think this has negative consequences as is.

It is at 1Gi now arbitrarily because, like you said, for out given storage class it has no effect.  When we move to a different storage class I think things will have to change regardless (ex: `storageClassName: microk8s-hostpath` will change, and we probably need user input about storage size rather than just a fixed amount).  So to me, whatever we choose here has no effect on anything :D  

But for that reason, I've changed it haha

kimwnasptd (approved):

Issue #39, opened 2024-03-04
Add option to `create-notebook` to use GPUs


ca-scribner:
### Why it needs to get done

Based on the [dss spec](https://docs.google.com/document/d/1v0wU8TBb-ui6RcjGRbLwA6XFZnWbuiVb2TPmEgwKejY/edit), we need notebooks that support GPUs.  #31 implements the `create-notebook` command, but it does not include an option to enable a GPU.  

### What needs to get done

1. add an interface to `create-notebook` to enable gpus (`--gpu` flag?  or something more detailed?) 

### When is the task considered done

1. `create-notebook` allows creation of notebooks with GPUs

NohaIhab:
[Note](https://github.com/canonical/data-science-stack/pull/43#discussion_r1552994370) from @misohu on the implementation details of this task

misohu:
We will get rid of the `--gpu` `--no-gpu` flags for now. The Pods have access to GPU by default after enabling the gpu addon and running correct image with drivers e.g. pytorch-cuda-full.

Issue #37, opened 2024-02-28
Improve how dss manages its kubeconfig file


ca-scribner:
### Why it needs to get done

#29 initializes dss on a given microk8s cluster, specifying the cluster via a [kubeconfig file passed by its filepath](https://github.com/canonical/data-science-stack/pull/29/files#diff-2f57dc654b74f69ba83e508cc30ef64d0c75f5d95664138c59e693663019d467R18).  This works fine for an initial implementation, but extending this to other commands will be frustrating because users will have to provide `kubeconfig` **for every call to the `dss` CLI**.  We need to implement a better way of remembering the kubeconfig file between command executions to maintain a positive user experience.  

### What needs to get done

Implement something to remember kubeconfig files between calls, for example maybe keeping a local copy of the most recent file provided.  This method must work both for our testing in this repo as well as when implemented as a snap.

### When is the task considered done

all `dss` commands have a way of remembering the kubeconfig between calls.

ca-scribner:
Another kinda edge case for why updating a kubeconfig could be important is if a user doesn't realize they can specify `--kubeconfig` during their first setup:
```bash
# user initializes
KUBECONFIG=~/.kube/config dss initialize

# user doesn't want to keep retyping KUBECONFIG, but can never recover...
KUBECONFIG=~/.kube/config dss status

# This will fail forever
dss status
```

But again of course the user could just update the file manually, and our help text could tell them about that

ca-scribner:
Minor update - #113 implements the above logic **and removes the personal-files plug that let dss access the user's `~/.dss/config` location**.  Strictly confined snaps always have access to their own "home" directory (ex: `/home/user/snap/data-science-stack/x1`), so the snap will now always cache the kubeconfig to that location rather than one in the user's home directory.  If a user wanted to edit the file manually, they can still do so from that snap dir.

Issue #34, opened 2024-02-26
Implement DSS `purge` command


DnPlas:
### Why it needs to get done

The `remove-components` command removes all Kubernetes components connected to DSS. Internally, this involves the removal of the entire `dss` Kubernetes namespace. Importantly, this command does not uninstall the DSS snap or the Python library. It proves useful for reinitialization without the necessity to remove the snap. It's important to note that PVC for notebooks will be removed but data will stay in microk8’s [hostpath folder](https://microk8s.io/docs/addon-hostpath-storage#Customize%20directory%20used%20for%20PersistentVolume).

```
dss remove-components
```

### What needs to get done

1. Implement `remove components` command which deploys aforementioned components.
2. Implement integration test with Microk8s

### When is the task considered done

1. Command is implemented
2. Tests are passing

misohu:
This command has been renamed to `dss purge` more information in [this](https://docs.google.com/document/d/1v0wU8TBb-ui6RcjGRbLwA6XFZnWbuiVb2TPmEgwKejY/edit#heading=h.vzt2qdtc70ug) document

misohu:
After discussion with UX team the `purge` command should  be blocking meaning this command can only finish after the components are completely removed from the cluster. This command should not have a timeout.

PR #43, opened 2024-03-07
implement `dss create-notebook` command

### LINKED ISSUE #31: Implement DSS `create-notebook` command

ca-scribner:
Adds the `create-notebook` command as well as unit and integration tests.

Closes #31.

### DIFF
--- a/src/dss/create_notebook.py
+++ b/src/dss/create_notebook.py
@@ -13,7 +13,7 @@ from dss.config import (
     NOTEBOOK_PVC_NAME,
 )
 from dss.logger import setup_logger
-from dss.utils import wait_for_deployment_ready, _get_notebook_url
+from dss.utils import wait_for_deployment_ready, get_notebook_url, get_mlflow_tracking_uri
 
 # Set up logger
 logger = setup_logger("logs/dss.log")
@@ -35,19 +35,14 @@ def create_notebook(name: str, image: str, lightkube_client: Client) -> None:
         Path(__file__).parent, MANIFEST_TEMPLATES_LOCATION, "notebook_deployment.yaml.j2"
     )
 
-    context = {
-        "notebook_name": name,
-        "namespace": DSS_NAMESPACE,
-        "notebook_image": image,
-        "pvc_name": NOTEBOOK_PVC_NAME,
-    }
+    config = _get_notebook_config(image, name)
 
     # Initialize KubernetesResourceHandler
     k8s_resource_handler = KubernetesResourceHandler(
         field_manager=FIELD_MANAGER,
         labels=DSS_CLI_MANAGER_LABELS,
         template_files=[manifests_file],
-        context=context,
+        context=config,
         resource_types={Deployment, Service},
         lightkube_client=lightkube_client,
     )
@@ -63,5 +58,19 @@ def create_notebook(name: str, image: str, lightkube_client: Client) -> None:
         k8s_resource_handler.delete()
         return
 
-    notebook_url = _get_notebook_url(name, DSS_NAMESPACE, lightkube_client)
+    notebook_url = get_notebook_url(name, DSS_NAMESPACE, lightkube_client)
     logger.info(f"Access the notebook at {notebook_url}.")
+
+
+def _get_notebook_config(image, name):
+    mlflow_tracking_uri = get_mlflow_tracking_uri()
+    context = {
+        "mlflow_tracking_uri": mlflow_tracking_uri,
+        "notebook_name": name,
+        "namespace": DSS_NAMESPACE,
+        "notebook_image": image,
+        "pvc_name": NOTEBOOK_PVC_NAME,
+    }
+    return context
+
+
--- a/src/dss/create_notebook.py
+++ b/src/dss/create_notebook.py
@@ -1,0 +1,67 @@
+from pathlib import Path
+
+from charmed_kubeflow_chisme.kubernetes import KubernetesResourceHandler
+from lightkube import Client
+from lightkube.resources.apps_v1 import Deployment
+from lightkube.resources.core_v1 import Service
+
+from dss.config import (
+    DSS_CLI_MANAGER_LABELS,
+    DSS_NAMESPACE,
+    FIELD_MANAGER,
+    MANIFEST_TEMPLATES_LOCATION,
+    NOTEBOOK_PVC_NAME,
+)
+from dss.logger import setup_logger
+from dss.utils import wait_for_deployment_ready, _get_notebook_url
+
+# Set up logger
+logger = setup_logger("logs/dss.log")
+
+
+def create_notebook(name: str, image: str, lightkube_client: Client) -> None:
+    """
+    Creates a Notebook server on the Kubernetes cluster.
+
+    Args:
+        name (str): The name of the notebook server.
+        image (str): The image used for the notebook server.
+        lightkube_client (Client): The Kubernetes client.
+
+    Returns:
+        None
+    """
+    manifests_file = Path(
+        Path(__file__).parent, MANIFEST_TEMPLATES_LOCATION, "notebook_deployment.yaml.j2"
+    )
+
+    context = {
+        "notebook_name": name,
+        "namespace": DSS_NAMESPACE,
+        "notebook_image": image,
+        "pvc_name": NOTEBOOK_PVC_NAME,
+    }
+
+    # Initialize KubernetesResourceHandler
+    k8s_resource_handler = KubernetesResourceHandler(
+        field_manager=FIELD_MANAGER,
+        labels=DSS_CLI_MANAGER_LABELS,
+        template_files=[manifests_file],
+        context=context,
+        resource_types={Deployment, Service},
+        lightkube_client=lightkube_client,
+    )
+
+    try:
+        k8s_resource_handler.apply()
+
+        wait_for_deployment_ready(lightkube_client, namespace=DSS_NAMESPACE, deployment_name=name)
+
+        logger.info("Notebook created.")
+    except TimeoutError as err:
+        logger.error(f"{str(err)}\nCleaning up the resources just created.")
+        k8s_resource_handler.delete()
+        return
+
+    notebook_url = _get_notebook_url(name, DSS_NAMESPACE, lightkube_client)
+    logger.info(f"Access the notebook at {notebook_url}.")
--- a/src/dss/create_notebook.py
+++ b/src/dss/create_notebook.py
@@ -13,7 +13,7 @@ from dss.config import (
     NOTEBOOK_PVC_NAME,
 )
 from dss.logger import setup_logger
-from dss.utils import wait_for_deployment_ready, get_notebook_url, get_mlflow_tracking_uri
+from dss.utils import get_mlflow_tracking_uri, get_notebook_url, wait_for_deployment_ready
 
 # Set up logger
 logger = setup_logger("logs/dss.log")
@@ -52,7 +52,7 @@ def create_notebook(name: str, image: str, lightkube_client: Client) -> None:
 
         wait_for_deployment_ready(lightkube_client, namespace=DSS_NAMESPACE, deployment_name=name)
 
-        logger.info("Notebook created.")
+        logger.info(f"Notebook {name} created.")
     except TimeoutError as err:
         logger.error(f"{str(err)}\nCleaning up the resources just created.")
         k8s_resource_handler.delete()
@@ -72,5 +72,3 @@ def _get_notebook_config(image, name):
         "pvc_name": NOTEBOOK_PVC_NAME,
     }
     return context
-
-
--- a/src/dss/main.py
+++ b/src/dss/main.py
@@ -1,5 +1,6 @@
 import click
 
+from dss.create_notebook import create_notebook
 from dss.initialize import initialize
 from dss.logger import setup_logger
 from dss.utils import KUBECONFIG_DEFAULT, get_default_kubeconfig, get_lightkube_client
@@ -28,6 +29,32 @@ def initialize_command(kubeconfig: str) -> None:
     lightkube_client = get_lightkube_client(kubeconfig)
 
     initialize(lightkube_client=lightkube_client)
+
+
+@main.command(name="create-notebook")
+@click.option(
+    "--name",
+    help="Name given to the notebook being created.",
+)
+@click.option(
+    "--image",
+    # TODO: Add default images
+    help="The image used for the notebook server.",
+)
+@click.option(
+    "--kubeconfig",
+    help=f"Path to a Kubernetes config file. Defaults to the value of the KUBECONFIG environment variable, else to '{KUBECONFIG_DEFAULT}'.",  # noqa E501
+)
+def create_notebook_command(name: str, image: str, kubeconfig: str) -> None:
+    """
+    Create a Notebook server.
+    """
+    logger.info("Executing create-notebook command")
+
+    kubeconfig = get_default_kubeconfig(kubeconfig)
+    lightkube_client = get_lightkube_client(kubeconfig)
+
+    create_notebook(name=name, image=image, lightkube_client=lightkube_client)
 
 
 if __name__ == "__main__":
--- a/src/dss/manifest_templates/notebook_deployment.yaml.j2
+++ b/src/dss/manifest_templates/notebook_deployment.yaml.j2
@@ -15,9 +15,6 @@ spec:
   template:
     metadata:
       labels:
-        # TODO: Need to integrate the things that these pod defaults used to represent
-{#        access-minio: "true"#}
-{#        mlflow-server-minio: "true"#}
         notebook-name: {{ notebook_name }}
         # TODO
         selector: TBD
@@ -26,28 +23,20 @@ spec:
         - env:
           - name: NB_PREFIX
             value: /notebook/{{ namespace }}/{{ notebook_name }}
+          - name: MLFLOW_TRACKING_URI
+            value: {{ mlflow_tracking_uri }}
           image: {{ notebook_image }}
           imagePullPolicy: IfNotPresent
-          # TODO
           name: {{ notebook_name }}
           ports:
           - containerPort: 8888
             name: notebook-port
             protocol: TCP
-          # TODO: add resource limits
-{#          resources:#}
-{#            limits:#}
-{#              cpu: 2400m#}
-{#              memory: 5153960755200m#}
-{#            requests:#}
-{#              cpu: "2"#}
-{#              memory: 4Gi#}
           terminationMessagePath: /dev/termination-log
           terminationMessagePolicy: File
           volumeMounts:
           - mountPath: /dev/shm
             name: dshm
-          # TODO
           - mountPath: /home/jovyan
             name: home-volume
           workingDir: /home/jovyan
@@ -58,7 +47,6 @@ spec:
         name: dshm
       - name: home-volume
         persistentVolumeClaim:
-          # TODO
           claimName: {{ pvc_name }}
 ---
 apiVersion: v1
@@ -76,12 +64,3 @@ spec:
     - protocol: TCP
       port: 80
       targetPort: 8888
-
-
-
-
-
-# Create pvc (shared)
-# * set owner references to the statefulset?  or maybe not, we probably dont want it deleted
-# Create statefulset
-# Create service (needed? do we do this for mlflow?)  (I guess it means ip doesn't change if pod restarts for some reason)
--- a/src/dss/manifest_templates/notebook_deployment.yaml.j2
+++ b/src/dss/manifest_templates/notebook_deployment.yaml.j2
@@ -1,0 +1,87 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ notebook_name }}
+  namespace: {{ namespace }}
+  labels:
+    # TODO
+    tbd: TBD
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      # TODO
+      selector: TBD
+  template:
+    metadata:
+      labels:
+        # TODO: Need to integrate the things that these pod defaults used to represent
+{#        access-minio: "true"#}
+{#        mlflow-server-minio: "true"#}
+        notebook-name: {{ notebook_name }}
+        # TODO
+        selector: TBD
+    spec:
+      containers:
+        - env:
+          - name: NB_PREFIX
+            value: /notebook/{{ namespace }}/{{ notebook_name }}
+          image: {{ notebook_image }}
+          imagePullPolicy: IfNotPresent
+          # TODO
+          name: {{ notebook_name }}
+          ports:
+          - containerPort: 8888
+            name: notebook-port
+            protocol: TCP
+          # TODO: add resource limits
+{#          resources:#}
+{#            limits:#}
+{#              cpu: 2400m#}
+{#              memory: 5153960755200m#}
+{#            requests:#}
+{#              cpu: "2"#}
+{#              memory: 4Gi#}
+          terminationMessagePath: /dev/termination-log
+          terminationMessagePolicy: File
+          volumeMounts:
+          - mountPath: /dev/shm
+            name: dshm
+          # TODO
+          - mountPath: /home/jovyan
+            name: home-volume
+          workingDir: /home/jovyan
+      terminationGracePeriodSeconds: 30
+      volumes:
+      - emptyDir:
+          medium: Memory
+        name: dshm
+      - name: home-volume
+        persistentVolumeClaim:
+          # TODO
+          claimName: {{ pvc_name }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ notebook_name }}
+  namespace: {{ namespace }}
+  labels:
+    # TODO
+spec:
+  selector:
+    # TODO
+    selector: TBD
+  ports:
+    - protocol: TCP
+      port: 80
+      targetPort: 8888
+
+
+
+
+
+# Create pvc (shared)
+# * set owner references to the statefulset?  or maybe not, we probably dont want it deleted
+# Create statefulset
+# Create service (needed? do we do this for mlflow?)  (I guess it means ip doesn't change if pod restarts for some reason)
--- a/src/dss/utils.py
+++ b/src/dss/utils.py
@@ -4,7 +4,9 @@ from typing import Optional
 
 from lightkube import Client, KubeConfig
 from lightkube.resources.apps_v1 import Deployment
+from lightkube.resources.core_v1 import Service
 
+from dss.config import DSS_NAMESPACE
 from dss.logger import setup_logger
 
 # Set up logger
@@ -13,9 +15,6 @@ logger = setup_logger("logs/dss.log")
 # Name for the environment variable storing kubeconfig
 KUBECONFIG_ENV_VAR = "DSS_KUBECONFIG"
 KUBECONFIG_DEFAULT = "./kubeconfig"
-
-# Labels applied to any Kubernetes objects managed by the DSS CLI
-DSS_CLI_MANAGER_LABELS = {"app.kubernetes.io/managed-by": "dss-cli"}
 
 
 def wait_for_deployment_ready(
@@ -84,3 +83,17 @@ def get_lightkube_client(kubeconfig: Optional[str] = None):
     kubeconfig = KubeConfig.from_file(kubeconfig)
     lightkube_client = Client(config=kubeconfig)
     return lightkube_client
+
+
+def _get_notebook_url(name: str, namespace: str, lightkube_client: Client) -> str:
+    """
+    Returns the URL of the notebook server given the name of the server.
+
+    Assumes the following conventions:
+    * the notebook server is exposed by a service of the same name
+    * the notebook service exposes the notebook on the first port in the service
+    """
+    service = lightkube_client.get(Service, namespace=DSS_NAMESPACE, name=name)
+    ip = service.spec.clusterIP
+    port = service.spec.ports[0].port
+    return f"http://{ip}:{port}/notebook/{namespace}/{name}/lab"
--- a/src/dss/utils.py
+++ b/src/dss/utils.py
@@ -14,6 +14,9 @@ logger = setup_logger("logs/dss.log")
 KUBECONFIG_ENV_VAR = "DSS_KUBECONFIG"
 KUBECONFIG_DEFAULT = "./kubeconfig"
 
+# Labels applied to any Kubernetes objects managed by the DSS CLI
+DSS_CLI_MANAGER_LABELS = {"app.kubernetes.io/managed-by": "dss-cli"}
+
 
 def wait_for_deployment_ready(
     client: Client,
--- a/src/dss/utils.py
+++ b/src/dss/utils.py
@@ -6,7 +6,7 @@ from lightkube import Client, KubeConfig
 from lightkube.resources.apps_v1 import Deployment
 from lightkube.resources.core_v1 import Service
 
-from dss.config import DSS_NAMESPACE
+from dss.config import DSS_NAMESPACE, MLFLOW_DEPLOYMENT_NAME
 from dss.logger import setup_logger
 
 # Set up logger
@@ -85,7 +85,11 @@ def get_lightkube_client(kubeconfig: Optional[str] = None):
     return lightkube_client
 
 
-def _get_notebook_url(name: str, namespace: str, lightkube_client: Client) -> str:
+def get_mlflow_tracking_uri() -> str:
+    return f"http://{MLFLOW_DEPLOYMENT_NAME}.{DSS_NAMESPACE}.svc.cluster.local:5000"
+
+
+def get_notebook_url(name: str, namespace: str, lightkube_client: Client) -> str:
     """
     Returns the URL of the notebook server given the name of the server.
 
--- a/tests/integration/test_dss.py
+++ b/tests/integration/test_dss.py
@@ -5,7 +5,7 @@ from charmed_kubeflow_chisme.kubernetes import KubernetesResourceHandler
 from lightkube.resources.apps_v1 import Deployment
 from lightkube.resources.core_v1 import Namespace, PersistentVolumeClaim, Service
 
-from dss.config import DSS_CLI_MANAGER_LABELS
+from dss.config import DSS_CLI_MANAGER_LABELS, FIELD_MANAGER
 
 
 def test_initialize_creates_dss(cleanup_after_initialize) -> None:
@@ -61,7 +61,7 @@ def cleanup_after_initialize():
     yield
 
     k8s_resource_handler = KubernetesResourceHandler(
-        field_manager="dss",
+        field_manager=FIELD_MANAGER,
         labels=DSS_CLI_MANAGER_LABELS,
         template_files=[],
         context={},
--- a/tests/integration/test_dss.py
+++ b/tests/integration/test_dss.py
@@ -1,14 +1,15 @@
 import subprocess
 
+import lightkube
 import pytest
 from charmed_kubeflow_chisme.kubernetes import KubernetesResourceHandler
 from lightkube.resources.apps_v1 import Deployment
 from lightkube.resources.core_v1 import Namespace, PersistentVolumeClaim, Service
 
-from dss.config import DSS_CLI_MANAGER_LABELS, FIELD_MANAGER
+from dss.config import DSS_CLI_MANAGER_LABELS, DSS_NAMESPACE, FIELD_MANAGER
 
 
-def test_initialize_creates_dss(cleanup_after_initialize) -> None:
+def test_initialize_creates_dss(cleanup_after_tests) -> None:
     """
     Integration test to verify if the initialize command creates the 'dss' namespace and
     the 'mlflow' deployment is active in the 'dss' namespace.
@@ -28,13 +29,13 @@ def test_initialize_creates_dss(cleanup_after_initialize) -> None:
 
     # Check if the dss namespace exists using kubectl
     kubectl_result = subprocess.run(
-        ["kubectl", "get", "namespace", "dss"], capture_output=True, text=True
+        ["kubectl", "get", "namespace", DSS_NAMESPACE], capture_output=True, text=True
     )
-    assert "dss" in kubectl_result.stdout
+    assert DSS_NAMESPACE in kubectl_result.stdout
 
     # Check if the mlflow-deployment deployment is active in the dss namespace
     kubectl_result = subprocess.run(
-        ["kubectl", "get", "deployment", "mlflow", "-n", "dss"],
+        ["kubectl", "get", "deployment", "mlflow", "-n", DSS_NAMESPACE],
         capture_output=True,
         text=True,
     )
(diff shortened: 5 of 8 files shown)


src/dss/main.py
```diff
@@ -30,5 +31,40 @@ def initialize_command(kubeconfig: str) -> None:
...
+- kubeflownotebookswg/jupyter-tensorflow-cuda-full:v1.8.0
+"""
+
+
+@main.command(name="create-notebook")
+@click.option(
+    "--name",
+    help="Name given to the notebook being created.",
+)
+@click.option(
```
DnPlas, line 49: ?

```
@main.command(name="create-notebook")
@click.option(
```

tests/integration/test_dss.py
```diff
@@ -45,23 +46,63 @@ def test_initialize_creates_dss(cleanup_after_initialize) -> None:
...
+    result = subprocess.run(
+        [
+            DSS_NAMESPACE,
+            "create-notebook",
+            "--name",
+            notebook_name,
+            "--image",
+            notebook_image,
+            "--kubeconfig",
+            kubeconfig_file,
```
DnPlas, line 77: Why are we passing kube config on `create-notebook`?

tests/integration/test_dss.py
```diff
@@ -45,23 +46,63 @@ def test_initialize_creates_dss(cleanup_after_initialize) -> None:
...
+        ],
+        capture_output=True,
+        text=True,
+        timeout=60 * 4,
+    )
+
+    # Check if the command executed successfully
+    print(result.stdout)
```
DnPlas, line 85: I believe `prints` are lost in integration checks. Should we log it instead?

DnPlas (commented):
Thanks @ca-scribner, few comments, I would like to review the command itself, but I'm having some issues even initialising `dss`. I have filed an [issue](https://github.com/canonical/data-science-stack/issues/44) commenting these things. Let me know if you have also ran into any of this so I can also test it manually.

src/dss/create_notebook.py
```diff
@@ -0,0 +1,74 @@
...
+        name (str): The name of the notebook server.
+        image (str): The image used for the notebook server.
+        lightkube_client (Client): The Kubernetes client.
+
+    Returns:
+        None
+    """
+    manifests_file = Path(
+        Path(__file__).parent, MANIFEST_TEMPLATES_LOCATION, "notebook_deployment.yaml.j2"
+    )
```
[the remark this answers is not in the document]
ca-scribner: to make sure I understand, do you mean in `config.py` we have things like
```python
NOTEBOOK_DEPLOYMENT_MANIFEST = Path(Path(__file__) ... "notebook_deployment.yaml.j2")
MLFLOW_DEPLOYMENT_MANIFEST = Path(Path(__file__) ... "mlflow_deployment.yaml.j2")

```
?

I'm cool with that.  My only worry is how `__file__` will work in that case.  Would need to test

src/dss/main.py
ca-scribner: I'm not sure what you're asking, can you elaborate?

tests/integration/test_dss.py
ca-scribner: dss has no "save a kubeconfig for later" functionality yet, so atm we have to pass it every time.

src/dss/manifest_templates/notebook_deployment.yaml.j2
```diff
@@ -0,0 +1,66 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ notebook_name }}
+  namespace: {{ namespace }}
+  labels:
+    # TODO
+    tbd: TBD
```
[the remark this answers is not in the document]
ca-scribner: Fixed these - now there's more meaningful labels

tests/integration/test_dss.py
ca-scribner: actually, both `print` and `log` are filtered out by pytest by default.  But specifying `pytest -s` (which is done in `tox.ini`) tells pytest to make the output of both visible.

We can still log though instead of print.  I don't know which is better for tests

src/dss/create_notebook.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    Creates a Notebook server on the Kubernetes cluster.
+
+    Args:
+        name (str): The name of the notebook server.
+        image (str): The image used for the notebook server.
+        lightkube_client (Client): The Kubernetes client.
+
+    Returns:
+        None
```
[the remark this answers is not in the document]
ca-scribner: removed

src/dss/create_notebook.py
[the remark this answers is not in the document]
DnPlas: Yeah, I meant exactly that. We would need to change slightly the path, but I think we can sort it out.

src/dss/main.py
DnPlas: Oh, sorry it was not clear. Aren't we missing the `@main.command(name="create-notebook")` decorator?

tests/integration/test_dss.py
DnPlas: I see. It's a minor thing and I will leave it up to you, just ensure the logs are actually printed (if you have to modify the tox, let's do it).

src/dss/create_notebook.py
```diff
@@ -0,0 +1,71 @@
...
+        field_manager=FIELD_MANAGER,
+        labels=DSS_CLI_MANAGER_LABELS,
+        template_files=[manifests_file],
+        context=config,
+        resource_types={Deployment, Service},
+        lightkube_client=lightkube_client,
+    )
+
+    try:
+        k8s_resource_handler.apply()
```
DnPlas, line 48: Should we handle`lightkube` errors? I think that if for any reason the notebook server deployment cannot be applied, it will throw an error instead of a timeout, which is not handled.

Also, what about the possibility of a person creating a notebook server with the same name, but different image, do you think applying will override the existing server? Should we have an imperative approach here to avoid modifying existing deployments?

src/dss/main.py
```diff
@@ -30,5 +31,40 @@ def initialize_command(kubeconfig: str) -> None:
     initialize(lightkube_client=lightkube_client)
 
 
+CREATE_NOTEBOOK_IMAGE_HELP = """
```
DnPlas, line 34: We are going to change this soon (when we have our rocks) could we have either a tracking issue or a comment that says these images are not the final ones that will be printed in the help?

src/dss/create_notebook.py
```diff
@@ -0,0 +1,71 @@
...
+
+        wait_for_deployment_ready(lightkube_client, namespace=DSS_NAMESPACE, deployment_name=name)
+
+        logger.info(f"Notebook {name} created.")
+    except TimeoutError as err:
+        logger.error(f"{str(err)}\nCleaning up the resources just created.")
+        k8s_resource_handler.delete()
```
DnPlas, line 55: I noticed that if the `create-notebook` command times out, it deletes everything in the `dss` namespace. The namespace is left intact, though. Could it be that we have the same labels everywhere and the `delete()` method catches it and just wipes out everything?

tests/integration/test_dss.py
```diff
@@ -45,23 +46,63 @@ def test_initialize_creates_dss(cleanup_after_initialize) -> None:
...
+    """
+    # TODO: is there a better way to initialize this?  Maybe an optional argument to the test?
+    kubeconfig_file = "~/.kube/config"
+    kubeconfig = lightkube.KubeConfig.from_file(kubeconfig_file)
+    lightkube_client = lightkube.Client(kubeconfig)
+
+    notebook_name = "test-notebook"
+    notebook_image = "kubeflownotebookswg/jupyter-scipy:v1.8.0"
+
+    result = subprocess.run(
```
DnPlas, line 68: Please add a `FIXME` comment that says we need to remove the `--kubeconfig` option after fixing #37

DnPlas (commented):
Hey @ca-scribner, some comments from #44 are still applicable (specially the manifest files one), but I was able to initialise and create a notebook with this python package. 

I have left some comments, but I think it is looking good overall.

src/dss/manifest_templates/notebook_deployment.yaml.j2
misohu, line 1: We should add this file and all the files in manifest_templates to the dss package 


this entry in setup.cfg was working for me.
```
[options.package_data]
dss = manifest_templates/*
```

ca-scribner: I swear I tried this, but I'll try it again.  Maybe I got something wrong before

src/dss/main.py
ca-scribner: oh I think that's [up here](https://github.com/canonical/data-science-stack/pull/43/files/627aff24391098bdb7a0506f3b11f173735eae7c#diff-2f57dc654b74f69ba83e508cc30ef64d0c75f5d95664138c59e693663019d467R44)

src/dss/main.py
ca-scribner: good call - captured as #57

tests/integration/test_dss.py
ca-scribner: Is this request just for here (the tests) or to remove the option from the command entirely?  I've added a `FIXME` in `main.py` for now with a note about how tests should change too.  Did I interpret this right?

src/dss/create_notebook.py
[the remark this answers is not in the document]
ca-scribner: I tested to make sure this would work and I think it does.  If we write in `config.py`:
```python
MANIFEST_TEMPLATES_LOCATION = Path("./manifest_templates")
ABSOLUTE_MANIFEST_TEMPLATES_LOCATION = Path(__file__).parent / MANIFEST_TEMPLATES_LOCATION
```

then the `__file__` in that call is relative to `config.py`, not `create_notebook.py` or whoever is importing `config.py`.  But I think that's ok.

Where this is following the same pattern as the other commands, do you mind if we implement this change separate to this PR and do it for all commands at once?  We'd have the same change for `initialize.py` and probably what Noha is implementing.

src/dss/create_notebook.py
ca-scribner: (I'm not sure deleting anything is the right choice if there's a timeout as it makes debugging hard.  But ignoring that separate conversation for now and assuming we do want to delete things on a timeout...) 

This is a good catch, and turns out its broken more than just this way :D

Yes the issue here is that the `DSS_CLI_MANAGER_LABELS` labels are used for anything the manager did (both from `initialize` and `create-notebook`).  So the `krh.delete()` here deleted all the Deployment and Services from both `initialize` and `create-notebook`.  The namespace (and pvc) were not deleted because the notebook's KRH is scoped only to deployments and services (it will never delete objects of other types).  

Another thing that could have gone wrong and I missed was this can happen:
1) `dss create-notebook notebook1` 
    1) assume this succeeds and we now have `notebook1`
2) `dss create-notebook notebook2`
    1) assume this times out, triggering deletion

On timeout, (2.i) will delete `notebook2` **and** `notebook1` (...and mlflow)

ca-scribner: I think the question here is what functionality do we need? We should have an easy way to:
1. delete all dss resources
2. delete a single notebook and its associated resources (`Service`)

I think labels are pretty good for (1).  The remove commands for dss could look for our `DSS_CLI_MANAGER_LABELS` in any object type we might create.  

For (2), we should be using whatever #34 / #51 do.  Otherwise we're just creating a second system of removal.  So maybe in this PR we should just log that something timed out, and then add the deletion in #51.  wdyt?

src/dss/manifest_templates/notebook_deployment.yaml.j2
ca-scribner: Seemed to work when I created a new venv.  Pushed the change, @DnPlas does it work for you?

src/dss/create_notebook.py
ca-scribner: Added the check in create_notebook

misohu:
We might need to postpone this PR until this issue is resolved https://github.com/canonical/data-science-stack/issues/58

misohu:
The dss create command has been discussed with UX team please check the DSS spec [here ](https://docs.google.com/document/d/1v0wU8TBb-ui6RcjGRbLwA6XFZnWbuiVb2TPmEgwKejY/edit#heading=h.wylzoqh1ph5w) for details about the error messages and parameters. 

I have also noticed @ca-scribner that we might have forgotten about gpu support for this command. More details are in the spec.

src/dss/create_notebook.py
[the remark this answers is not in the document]
NohaIhab: i filed issue https://github.com/canonical/data-science-stack/issues/61 to track this
@DnPlas resolve the conversation if you're okay with that

src/dss/manifest_templates/notebook_deployment.yaml.j2
DnPlas: It works for me, @misohu has added the change in #55, but [I have asked](https://github.com/canonical/data-science-stack/pull/55#discussion_r1535257865) him to send an isolated PR for it. This conversation can be reoslved.

src/dss/create_notebook.py
DnPlas: > Yeah I see what you mean about the `lightkube` errors - do you have an opinion on what we should do? The only thing I can think of is catching lightkube errors (ApiError?) and maybe raising a new error from that that adds us saying that something went wrong but we don't know what? I think we'd want to expose lightkube's traceback (since there's probably useful debugging info there) but we should add some value too

We could expose the traceback in a `debug` log so it helps when issues are filed. We'll need a mechanism for exposing debug lugs, a verbose mode if you will.
Regarding the message users will get, perhaps this is a question for the UX team, but for now I think it would be nice to show at least a msg that says there was an issue, the issue code, and a suggested fix(?).

DnPlas: > Since every DSS notebook uses the same PVC, I don't think there's much value in dss overwriting existing notebooks. Users can just remove and create them without losing their data. For now I'll add a check for an existing notebook of the same name, and warn the user if that happens. wdyt?

I think it is fine if we warn users for now as this is also a good question for the UX team. Let's follow up with them.

src/dss/create_notebook.py
DnPlas: > So maybe in this PR we should just log that something timed out, and then add the deletion in https://github.com/canonical/data-science-stack/pull/51. wdyt?

I think it is fine if we log that something timed out, but then we'll hit the case where the resource is not deleted and then after a while it could be up and ready without noticing. I'd say let's not remove, log the timeout, but warn people that the notebook server might be in the cluster and ways to check it (`dss list`)? It feels awkward, but at least it gives users a bit more visibility.

Kind of related:
While I agree that having the `dss create-notebook` command to output the notebook server address is really handy, and for knowing that address we need a timeout, I am wondering if the timeout is a good thing to have. First of all, users do not have control on the timeout time (as mentioned in #44), and even if they did, there are things that don't entirely depend on time: lightkube errors, memory, disk space, even gpus, internet connectivity, etc.
I know this was not in the spec, but perhaps a better approach would be to have something similar to what microk8s or juju does with status to show the notebook servers by name, if they are running or not, and their address. This way we avoid having to delete deployments and services, as users will have full access to the state of notebook servers.

If you feel this is out of scope of this PR, I suggest we file an issue to follow up on this.

tests/integration/test_dss.py
DnPlas: I meant remove from the command entirely, this just happens to be here (:

I do not see the FIXME, did I miss something?

src/dss/create_notebook.py
NohaIhab: > Since every DSS notebook uses the same PVC, I don't think there's much value in dss overwriting existing notebooks. Users can just remove and create them without losing their data. For now I'll add a check for an existing notebook of the same name, and warn the user if that happens. wdyt?

in the updated spec with the UX team's feedback:
```
Output upon name failure exists:
# fallback error
Failed to create notebook. Notebook <NAME> already exists.Please specify a different name.
```
Andrew's change is similar to that, just the error message needs a little tweaking.

tests/integration/test_dss.py
NohaIhab: @DnPlas I see the FIXME is in main.py L54

src/dss/create_notebook.py
NohaIhab: The deletion on timeout is indeed a problem when there's an unexpected error, then we will not be able to debug it.
For this PR, I will move on with @DnPlas 's suggestion to log the timeout, but warn people that the notebook server might be in the cluster to check it with `dss list`.

Whether we should have timeout or not is a discussion that we should have as a team and get feedback from the UX team on. I'd say even update the spec about this. cc: @misohu 
Let me file an issue to track this and let's discuss further there.

NohaIhab: filed #63 to discuss the timeout

orfeas-k:
If this is merged after #55, let's make sure it updates also tests accordingly like [here](https://github.com/canonical/data-science-stack/pull/55/files#diff-52f36670a957c0465b5d761dfd0164eb6538f17fbca1edcecee0e7da759dd48fR67)

tests/unit/test_create_notebook.py
```diff
@@ -54,20 +54,22 @@ def test_create_notebook_success(
...
-        # Mock wait_for_deployment_ready
-        with patch(
-            "dss.create_notebook.wait_for_deployment_ready"
-        ) as mock_wait_for_deployment_ready:
-            # Call the function to test
-            create_notebook(
-                name=notebook_name, image=notebook_image, lightkube_client=mock_client_instance
-            )
+    # Mock away the check on existing dss pvc
+    with patch("dss.create_notebook.does_dss_pvc_exist", return_value=True):
```
ca-scribner, line 58: Do we know what python version we want to require here?  If we say python>=3.10, we can [unnest these with blocks](https://stackoverflow.com/a/50958431)

tests/unit/test_utils.py
```diff
@@ -181,6 +187,39 @@ def test_wait_for_deployment_ready_timeout(mock_client: MagicMock, mock_logger:
...
+    pod.status.containerStatuses[0].state.waiting.reason = "ImagePullBackOff"
+    mock_client_instance.list.return_value = iter([pod])
+
+    # Call the function to test
+    with pytest.raises(ImagePullBackOffError) as exc_info:
+        wait_for_deployment_ready(
+            mock_client_instance,
+            namespace="test-namespace",
+            deployment_name="test-deployment",
+            timeout_seconds=5,
```
ca-scribner, line 213: (suggestion) I dont think it needs to be much longer

```suggestion
            timeout_seconds=2,
```

src/dss/utils.py
```diff
@@ -50,6 +62,21 @@ def wait_for_deployment_ready(
...
+                client.list(
+                    Pod,
+                    namespace=namespace,
+                    labels={"canonical.com/dss-notebook": deployment_name},
+                )
+            )[0]
+            reason = pod.status.containerStatuses[0].state.waiting.reason
+            if reason in ["ImagePullBackOff", "ErrImagePull"]:
+                raise ImagePullBackOffError(
+                    f"Failed to create Deployment {deployment_name} with {reason}"
```
ca-scribner, line 77: I think in this case we should still mention timeout.  It isn't that this failed because we're in ImagePullBackOff, its that we timed out and at the end we see we're still in that status.  Sometimes ImagePullBackOff is a transient thing, so I suggest we're clear that isn't what we think happened

```suggestion
                    f"Timed out trying to create Deployment {deployment_name} with {reason}"
```

ca-scribner (commented):
Added a few comments but overall looking good

setup.cfg
```diff
@@ -13,7 +13,7 @@ install_requires =
 include_package_data = True
 
 [options.package_data]
-dss = manifests.yaml
+dss = manifest_templates/*
```
misohu, line 16: This change was already merged to main please rebase.

src/dss/create_notebook.py
```diff
@@ -0,0 +1,124 @@
...
+    wait_for_deployment_ready,
+)
+
+NOTEBOOK_IMAGES_ALIASES = {
+    "pytorch": "charmedkubeflow/jupyter-pytorch-full:1.8.0-3058193",
+    # TODO: add the rest of the rocks once updated in https://github.com/canonical/kubeflow-rocks
+    # "pytorch-cuda": "charmedkubeflow/jupyter-pytorch-cuda-full:1.8.0-xxxxx",
+    # "tensorflow-cuda": "charmedkubeflow/jupyter-tensorflow-cuda-full:1.8.0-xxxx",
+    # "tensorflow": "charmedkubeflow/jupyter-tensorflow-full:1.8.0-xxxx"
+}
```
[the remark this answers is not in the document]
misohu: Agree with @ca-scribner  we should move it to config.

src/dss/main.py
```diff
@@ -30,5 +31,48 @@ def initialize_command(kubeconfig: str) -> None:
...
+    default="kubeflownotebookswg/jupyter-scipy:v1.8.0",
+    help=CREATE_NOTEBOOK_IMAGE_HELP,
+)
+# FIXME: Remove the kubeconfig param from the create-notebook command (and any tests) after
+#  https://github.com/canonical/data-science-stack/issues/37
+@click.option(
+    "--kubeconfig",
+    help=f"Path to a Kubernetes config file. Defaults to the value of the KUBECONFIG environment variable, else to '{KUBECONFIG_DEFAULT}'.",  # noqa E501
+)
+def create_notebook_command(name: str, image: str, kubeconfig: str) -> None:
```
misohu, line 60: I think we are missing functionality for `--gpu` and `--no-gpu` options which we mentioned in the doc here https://docs.google.com/document/d/1v0wU8TBb-ui6RcjGRbLwA6XFZnWbuiVb2TPmEgwKejY/edit#heading=h.wylzoqh1ph5w (check the full help message there is description of the options).

src/dss/main.py
```diff
@@ -30,5 +31,48 @@ def initialize_command(kubeconfig: str) -> None:
     initialize(lightkube_client=lightkube_client)
 
 
+CREATE_NOTEBOOK_IMAGE_HELP = """
+\b\nThe image used for the notebook server.  Suggested images:
+- kubeflownotebookswg/jupyter-scipy:v1.8.0 (default)
+- kubeflownotebookswg/jupyter-pytorch-full:v1.8.0
+- kubeflownotebookswg/jupyter-pytorch-cuda-full:v1.8.0
+- kubeflownotebookswg/jupyter-tensorflow-full:v1.8.0
+- kubeflownotebookswg/jupyter-tensorflow-cuda-full:v1.8.0
```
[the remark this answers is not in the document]
misohu: Agree with @ca-scribner IMO we can construct the whole help message in the config file and then export here.

src/dss/manifest_templates/notebook_deployment.yaml.j2
```diff
@@ -0,0 +1,67 @@
...
+      app.kubernetes.io/part-of: dss
+      canonical.com/dss-notebook: {{ notebook_name }}
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/part-of: dss
+        canonical.com/dss-notebook: {{ notebook_name }}
+    spec:
+      containers:
+        - env:
```
misohu, line 23: Comming back to the GPU options. When user speciffies `--gpu nvidia` in this manifest we will add resource section as shown in [this](https://github.com/canonical/data-science-stack/blob/main/poc/nvidia/manifests/notebook-nvidia.yaml#L45) example. Create notebook should also check if the gpu is enabled on the cluster by checking the node labels. If nvidia is enabled there will be dedicated label on the node and the resource will be present on the node. If you want to check the labels and resources on node you can deploy the nvidia poc from [here](https://github.com/canonical/data-science-stack/blob/main/poc/nvidia/deploy.sh). But you need a machine with nvidia card e.g Ec2 instance [list](https://docs.aws.amazon.com/dlami/latest/devguide/gpu.html).

src/dss/utils.py
misohu: I agree with the timeout but based on the UX team feedback we should avoid using tterms like Pod and Deployment when user is actually creating notebooks. We can have our own debug message in logger.debug with the pods and deployments but for the user error message we should rather avoid that. Moreover another UX recommendation is always to provide recommended next steps on errors if there are some (writing it here maybe it does not apply).

src/dss/utils.py
```diff
@@ -81,3 +113,62 @@ def get_lightkube_client(kubeconfig: Optional[str] = None):
     kubeconfig = KubeConfig.from_file(kubeconfig)
     lightkube_client = Client(config=kubeconfig)
     return lightkube_client
+
+
+def get_mlflow_tracking_uri() -> str:
```
misohu, line 118: Just a thought we are checking if PVC exists but we dont check if mlflow exists. We are by default always sending this url. Maybe we should also check for MLflow.

src/dss/main.py
```diff
@@ -30,5 +31,48 @@ def initialize_command(kubeconfig: str) -> None:
...
+
+@main.command(name="create-notebook")
+@click.argument(
+    "name",
+    required=True,
+)
+@click.option(
+    "--image",
+    default="kubeflownotebookswg/jupyter-scipy:v1.8.0",
+    help=CREATE_NOTEBOOK_IMAGE_HELP,
```
misohu, line 52: The help message should be as close as possible to the ux spec doc. Please change the summary plus add example section (I know click will always put it before options :)).

misohu (commented):
First itteration of review ... we need to add `--gpu` and `--no-gpu` options.

src/dss/utils.py
NohaIhab: what should we do if mlflow does not exist? not create the notebook? or create the notebook without the `MLFLOW_TRACKING_URI` env var?

misohu: I believe we dont expect a valid notebook without mlflow running ... I think we should not create notebook

src/dss/main.py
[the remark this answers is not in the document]
NohaIhab: I defined the message in the config file and imported it where needed

src/dss/create_notebook.py
```diff
@@ -64,23 +83,42 @@ def create_notebook(name: str, image: str, lightkube_client: Client) -> None:
...
+        logger.error(
+            f"Failed to create Notebook with error code {err.status.code}"
+            "   Check the debug logs for more details."
+        )
+        logger.debug(f"Failed to create Notebook {name} with error {err}")
     except TimeoutError as err:
-        logger.error(f"{str(err)}\nCleaning up the resources just created.")
-        k8s_resource_handler.delete()
+        logger.error(str(err))
+        logger.warn(f"Notebook {name} might be in the cluster. Check the status with `dss list`.")
```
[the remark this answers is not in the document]
NohaIhab: agreed, updated

src/dss/create_notebook.py
```diff
@@ -0,0 +1,124 @@
...
+            f"Failed to create Notebook {name}. Image {image} does not exist or is not accessible."
+        )
+        return
+    notebook_url = get_notebook_url(name, DSS_NAMESPACE, lightkube_client)
+    logger.info(f"Access the notebook at {notebook_url}.")
+
+
+def _get_notebook_config(image, name):
+    mlflow_tracking_uri = get_mlflow_tracking_uri()
+    full_image = _get_notebook_image_name(image)
```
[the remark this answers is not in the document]
NohaIhab: good point, I made some changes

src/dss/create_notebook.py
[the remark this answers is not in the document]
NohaIhab: moved

tests/unit/test_create_notebook.py
NohaIhab: updated, given that we agreed in matrix on python>=3.10

tests/unit/test_utils.py
NohaIhab: sgtm, updated

src/dss/utils.py
NohaIhab: I changed it to say timeout and logged this in the debug log, the error log prints `Notebook` not `Deployment`

setup.cfg
NohaIhab: merged

src/dss/main.py
NohaIhab: to be addressed in #39

src/dss/manifest_templates/notebook_deployment.yaml.j2
NohaIhab: to be addressed in #39 
I'll add a link to this comment there

src/dss/utils.py
NohaIhab: makes sense, I added the check alongside the PVC check

src/dss/main.py
NohaIhab: I added the example section, here's how it looks now:
```
dss create --help
Usage: dss create [OPTIONS] NAME

  Create a Jupyter notebook in DSS and connect it to MLflow. This command also
  outputs the URL to access the notebook on success.

  Examples:
      dss create my-notebook --image=pytorch
      dss create my-notebook --image=kubeflownotebookswg/jupyter-scipy:v1.8.0

Options:
  --image TEXT       The image used for the notebook server.
                     Recommended images:
                       - pytorch = charmedkubeflow/jupyter-pytorch-full:1.8.0-3058193
  --kubeconfig TEXT  Path to a Kubernetes config file. Defaults to the value
                     of the KUBECONFIG environment variable, else to
                     './kubeconfig'.
  --help             Show this message and exit.
```

orfeas-k: Modified this a bit 

```diff
-                       - pytorch = charmedkubeflow/jupyter-pytorch-full:1.8.0-3058193
+                       - pytorch: charmedkubeflow/jupyter-pytorch-full:1.8.0-3058193

```

src/dss/main.py
```diff
@@ -31,6 +33,47 @@ def initialize_command(kubeconfig: str) -> None:
...
+    help=f"Path to a Kubernetes config file. Defaults to the value of the KUBECONFIG environment variable, else to '{KUBECONFIG_DEFAULT}'.",  # noqa E501
+)
+def create_notebook_command(name: str, image: str, kubeconfig: str) -> None:
+    """Create a Jupyter notebook in DSS and connect it to MLflow.
+    This command also outputs the URL to access the notebook on success.
+
+    \b
+    Examples:
+        dss create my-notebook --image=pytorch
+        dss create my-notebook --image=kubeflownotebookswg/jupyter-scipy:v1.8.0
```
misohu, line 62: Help message should also provide list of recommended images. Check the ux [spec](https://docs.google.com/document/d/1v0wU8TBb-ui6RcjGRbLwA6XFZnWbuiVb2TPmEgwKejY/edit#heading=h.wylzoqh1ph5w) for refference.

src/dss/create_notebook.py
```diff
@@ -0,0 +1,129 @@
...
+
+    Args:
+        name (str): The name of the notebook server.
+        image (str): The image used for the notebook server.
+        lightkube_client (Client): The Kubernetes client.
+    """
+    if not does_dss_pvc_exist(lightkube_client) or not does_mlflow_deployment_exist(
+        lightkube_client
+    ):
+        logger.error(
```
misohu, line 45: We are little inconsistent with these messages .... in logs [here](https://github.com/canonical/data-science-stack/blob/main/src/dss/logs.py#L50) we write the first line as an error log while we write the hints as infos. Here we put everything in error log. The reasoning behind logs design was that info logs and error logs are stored in separate files for dss. With this in mind we are putting hints into info logs as when sb will we checking error logs they dont need the hints there.

src/dss/utils.py
```diff
@@ -81,3 +113,79 @@ def get_lightkube_client(kubeconfig: Optional[str] = None):
...
+    Returns the URL of the notebook server given the name of the server.
+
+    Assumes the following conventions:
+    * the notebook server is exposed by a service of the same name
+    * the notebook service exposes the notebook on the first port in the service
+    """
+    service = lightkube_client.get(Service, namespace=DSS_NAMESPACE, name=name)
+    ip = service.spec.clusterIP
+    port = service.spec.ports[0].port
+    return f"http://{ip}:{port}/notebook/{namespace}/{name}/lab"
```
misohu, line 133: This is incorrectl url the suffix /notebook/namespace/name/lab is not needed its not working with it, we need just the url.

tests/unit/test_create_notebook.py
misohu, line 1: What would be nice is to have an integration test which will create notebook and then it will run the script to create experiement in MLflow. e.g.

```
import mlflow
c = mlflow.MlflowClient()  # You should not need to specify the tracking_uri - it should be populated automatically
print(c.tracking_uri)  # should print the in-cluster url for dss's mlflow
c.create_experiment("test-experiment")  # Should succeed
```

The idea is to create notebook and then run kubectl exec with that script. Its little bit of a challenging task so we can maybe just create issue for that and face in separate task but nevertheless we will need to do it for complete tests in the future.

src/dss/utils.py
```diff
@@ -45,6 +62,21 @@ def wait_for_deployment_ready(
             logger.info(f"Deployment {deployment_name} in namespace {namespace} is ready")
             break
         elif time.time() - start_time >= timeout_seconds:
```
misohu, line 64: I vote for increating the timeout to at least 180 seconds ... sometimes I am running tests on smaller aws instances and the mlflow deployment on initialise is timing out ... I think 180 is reasonable.

tests/unit/test_create_notebook.py
misohu: or if there is a way to execute jupyter notebook from inside the pod ... we can git clone it and run.

src/dss/utils.py
```diff
@@ -81,3 +113,79 @@ def get_lightkube_client(kubeconfig: Optional[str] = None):
...
+
+def get_notebook_url(name: str, namespace: str, lightkube_client: Client) -> str:
+    """
+    Returns the URL of the notebook server given the name of the server.
+
+    Assumes the following conventions:
+    * the notebook server is exposed by a service of the same name
+    * the notebook service exposes the notebook on the first port in the service
+    """
+    service = lightkube_client.get(Service, namespace=DSS_NAMESPACE, name=name)
```
misohu, line 130: We are not handling case when this line  fails we are not catching it here nor we do it in the function where we use it.

src/dss/utils.py
```diff
@@ -81,3 +113,79 @@ def get_lightkube_client(kubeconfig: Optional[str] = None):
     kubeconfig = KubeConfig.from_file(kubeconfig)
     lightkube_client = Client(config=kubeconfig)
     return lightkube_client
+
+
+def get_mlflow_tracking_uri() -> str:
+    return f"http://{MLFLOW_DEPLOYMENT_NAME}.{DSS_NAMESPACE}.svc.cluster.local:5000"
+
+
+def get_notebook_url(name: str, namespace: str, lightkube_client: Client) -> str:
```
misohu, line 122: Can we rename this to get_service_url ... we can reuse it also for other services not just notebooks

src/dss/create_notebook.py
orfeas-k: good one, updated!

src/dss/utils.py
orfeas-k: Agreed, setting it as a default for both when creating a notebook and initializing DSS.

src/dss/utils.py
orfeas-k: Syncing on this with Michal, this is true but we 'll have to remove the `NB_PREFIX` env variable from the notebook deployment yaml, given that it's not needed and it's a leftover from kubeflow.

src/dss/utils.py
orfeas-k: Updated

tests/unit/test_create_notebook.py
orfeas-k: I 'm pro to file an issue and keep it as a separate task. Added a note in the daily to discuss.

src/dss/utils.py
orfeas-k: Added catching inside the function and taking that into consideration in the function above, lmk what you think.

src/dss/create_notebook.py
```diff
@@ -0,0 +1,135 @@
...
+        )
+        logger.debug(f"Timed out while trying to create Notebook {name} with error {err}.")
+        return
+    # Assumes that the notebook server is exposed by a service of the same name.
+    url = get_service_url(name, DSS_NAMESPACE, lightkube_client)
+    if url:
+        logger.info(f"Access the notebook at {url}.")
+
+
+def _get_notebook_config(image, name):
```
misohu, line 118: This function is missing type hints and docs strings

src/dss/create_notebook.py
```diff
@@ -0,0 +1,135 @@
...
+        "mlflow_tracking_uri": mlflow_tracking_uri,
+        "notebook_name": name,
+        "namespace": DSS_NAMESPACE,
+        "notebook_image": image,
+        "pvc_name": NOTEBOOK_PVC_NAME,
+    }
+    return context
+
+
+def _get_notebook_image_name(image) -> str:
```
misohu, line 130: Missing type hints

tests/unit/test_create_notebook.py
```diff
@@ -0,0 +1,297 @@
...
+            f"Timed out while trying to create Notebook {notebook_name}.\n"
+            f"Image {notebook_image} does not exist or is not accessible.\n"
+        )
+        mock_logger.info.assert_called_with(
+            "Note: You might want to use some of these recommended images:\n"
+            f"{RECOMMENDED_IMAGES_MESSAGE}"
+        )
+
+
+# def test_get_notebook_config() -> None:
```
misohu, line 281: Should we remove this ?

misohu (commented):
Small comments otherwise looking really good :)

tests/integration/test_dss.py
orfeas-k: Resolving since there is a fixme where Noha pointed out. Feel free to reopen if we need to add it in every place we use it. However, tests will fail when we remove it given that `kubeconfig` option won't be supported.

tests/unit/test_create_notebook.py
orfeas-k: Brought it back, I commented it out by accident.

misohu (approved):
LGTM good job @orfeas-k

Issue #28, opened 2024-02-12
Create a CI for publishing DSS snap


DnPlas:
### Why it needs to get done

To have an automated way of publishing DSS to the Snap store.

### What needs to get done

An automated workflow that can be used for releasing the DSS to the Snap Store on an agreed event (on demand, on merge, etc.) Please NOTE this task depends on #24, if we are able to leverage other solutions (like https://snapcraft.io/build) we may not need to build a workflow for this at all.

### When is the task considered done

Please NOTE this task depends on https://github.com/canonical/data-science-stack/issues/23, the description of this task will change accordingly.

misohu:
With this issue we are still aligning on name for snap in snapstore after that the snapstore should automatically build the snap from default branch to edge track. With regards to this I opened this issue https://github.com/canonical/data-science-stack/issues/56

misohu:
The snap name `data-science-stack` is reserved. I have requested 

1. alias dss for the snap https://forum.snapcraft.io/t/alias-request-for-data-science-stack/39515
2. change the ownership from me to Canonical https://forum.snapcraft.io/t/transfer-request-data-science-stack-to-canonical/39518

Finding: If you want to be part of the forum you need to create an account the Canonical one will not work.

PR #49, opened 2024-03-20
This snap is arch=any so lets only build on amd64

kenvandine:
No point in building for every arch on every commit using the snapcraft build service.


misohu (approved):
Approved ... this PR is connected with this issue  https://github.com/canonical/data-science-stack/issues/28

DnPlas (approved):

PR #51, opened 2024-03-21
feat: add remove notebook command

### LINKED ISSUE #32: Implement DSS `remove` command
### Why it needs to get done

The remove-notebook command removes a specified notebook using the lightkube library and prints "notebook removed" upon completion.

```
dss remove-notebook --name user-notebook
```

### What needs to get done

1. Implement `remove` command which deploys aforementioned components.

1. Command is implemented
2. Tests are passing

### LINKED ISSUE #34: Implement DSS `purge` command

NohaIhab:
closes #32


src/dss/remove_notebook.py
```diff
@@ -0,0 +1,32 @@
...
+    Removes a Notebook server from the Kubernetes cluster.
+
+    Args:
+        name (str): The name of the notebook server.
+        lightkube_client (Client): The Kubernetes client.
+    """
+
+    try:
+        lightkube_client.delete(res=Deployment, name=name, namespace=DSS_NAMESPACE)
+        lightkube_client.delete(res=Service, name=name, namespace=DSS_NAMESPACE)
```
ca-scribner, line 24: Just a drive-by comment on this that came up in #43 (specifically [here](https://github.com/canonical/data-science-stack/pull/43/files/627aff24391098bdb7a0506f3b11f173735eae7c#r1523105906))...  I realized create-notebook was creating its own deletion function when things timed out, which is not good.  

I wonder if we should use an `ownerReference` on the `Service` so that it will automatically get cleaned up if the notebook `Deployment` is deleted

src/dss/main.py
```diff
@@ -30,5 +31,28 @@ def initialize_command(kubeconfig: str) -> None:
     initialize(lightkube_client=lightkube_client)
 
 
+# FIXME: remove the `--kubeconfig`` option
+# after fixing https://github.com/canonical/data-science-stack/issues/37
+@main.command(name="remove-notebook")
+@click.option(
+    "--name",
```
misohu, line 38: I think the name of the notebook should be required. What if not provided ? Should it remove all notebooks ?

src/dss/remove_notebook.py
```diff
@@ -0,0 +1,32 @@
...
+    except ApiError as err:
+        if err.status.code == 404:
+            logger.warn("Failed to delete K8s resources not found. Ignoring remove-notebook.")
```
misohu, line 27: Maybe we can add some more helpful message recommendind what to do e.g. 

```
try dss list-notebooks to use the correct notebook name
```

misohu (commented):
Added small comments ... still I think we should adjust the integration test and merge this after we merge @ca-scribner PR https://github.com/canonical/data-science-stack/pull/43

src/dss/main.py
NohaIhab: good point

misohu:
We might need to postpone this PR until this issue is resolved https://github.com/canonical/data-science-stack/issues/58

src/dss/main.py
NohaIhab: done

src/dss/remove_notebook.py
NohaIhab: nice suggestion, done

ca-scribner:
@NohaIhab I think this actually is meant to close #32, not #34.  #34 is to remove everything (all notebooks, mlflow, etc) in one shot

src/dss/remove_notebook.py
ca-scribner: @kimwnasptd probably has a better sense of what is appropriate.  But I wouldn't block this PR on the choice.  If this PR gets to the point where everything else is reviewed and ready to go, lets merge with whatever you choose and we can always change it after.

src/dss/remove_notebook.py
```diff
@@ -0,0 +1,34 @@
...
+from dss.config import DSS_NAMESPACE
+from dss.logger import setup_logger
+
+# Set up logger
+logger = setup_logger("logs/dss.log")
+
+
+def remove_notebook(name: str, lightkube_client: Client) -> None:
+    """
+    Removes a Notebook server from the Kubernetes cluster.
```
ca-scribner, line 15: `Notebook` is a bit abstract - can you have a second paragraph that says specifically what will be removed (the `Deployment` and `Service` of name `name`)

tests/integration/test_dss.py
```diff
@@ -72,3 +117,28 @@ def cleanup_after_initialize():
     # Note that .delete() does not wait on the objects to be successfully deleted, so repeating
     # the tests quickly can still cause an issue
     k8s_resource_handler.delete()
+
+
+# FIXME: remove function when https://github.com/canonical/data-science-stack/pull/43 is merged
```
ca-scribner, line 122:
```suggestion
# FIXME: remove function and `tests/integration/deploy.yaml` when 
#  https://github.com/canonical/data-science-stack/pull/43 is merged
```

src/dss/remove_notebook.py
```diff
@@ -0,0 +1,34 @@
...
+
+    try:
+        lightkube_client.delete(res=Deployment, name=name, namespace=DSS_NAMESPACE)
+        lightkube_client.delete(res=Service, name=name, namespace=DSS_NAMESPACE)
+    except ApiError as err:
+        if err.status.code == 404:
+            logger.warn(
+                "Failed to delete K8s resources not found. Ignoring remove-notebook.\nTry dss list-notebooks to use the correct notebook name."  # noqa E501
+            )
+            return
```
ca-scribner, line 30: Should this count as a successful execution of `dss remove-notebook`?  As is, this warns and then completes successfully, so something like:
```
dss remove-notebook notARealNotebook && echo success
```
would show the warning and then echo "success"

imo, we should fail here rather than return and succeed (`exit(1)` maybe? or probably raise a custom exception that the `main.py` catches and `main.py` then does `exit(1)`).  Its a bit arbitrary, but I don't think we want to communicate that things went as we hoped.

For comparison, doing `kubectl delete namespace not-real` exits with error code 1

NohaIhab:
thanks @ca-scribner 
This is moving to blocked by #58, once it is unblocked we can reiterate on the comments.

misohu:
The dss remove command has been discussed with UX team please check the DSS spec [here ](https://docs.google.com/document/d/1v0wU8TBb-ui6RcjGRbLwA6XFZnWbuiVb2TPmEgwKejY/edit#heading=h.bvvmo5r2r72m) for details about the error messages and parameters.

src/dss/remove_notebook.py
NohaIhab: I think this applies to all the dss commands we've implemented so far, what we've done up till now is log errors rather than exit with error code. I will bring this up with the team so we can make a decision.

tests/integration/test_dss.py
NohaIhab: no need for this comment anymore

src/dss/remove_notebook.py
```diff
@@ -0,0 +1,42 @@
...
+        if err.status.code == 404:
+            logger.warn(
+                f"Failed to remove notebook. Notebook {name} does not exist. Run 'dss list' to check all notebooks."  # noqa E501
+            )
+            return
+        else:
+            logger.error(f"Failed to remove notebook {name}. Please try again.")
+            logger.info(
+                "You might want to run\n"
+                "  dss status      to check the current status\n"
```
misohu, line 37: Lets rewrite the info message here the same way we o for logs https://github.com/canonical/data-science-stack/blob/main/src/dss/logs.py#L54

src/dss/remove_notebook.py
NohaIhab: added

src/dss/remove_notebook.py
NohaIhab: good point, changed now

tests/integration/test_dss.py
```diff
@@ -130,6 +130,37 @@ def test_log_command(cleanup_after_initialize) -> None:
...
+        ]
+    )
+    assert result.returncode == 0
+
+    # Check if the notebook Deployment is not found in the namespace
+    with pytest.raises(ApiError):
+        lightkube_client.get(Deployment, name=NOTEBOOK_NAME, namespace=DSS_NAMESPACE)
+
+    # Check if the notebook Service is not found in the namespace
+    with pytest.raises(ApiError):
```
DnPlas, line 160: I think you want to check that the exception that gets raised has code 404, otherwise you could be hitting false positives.

src/dss/remove_notebook.py
DnPlas: I believe both the labels and the `ownerReference` will be helpful here. It looks like in this PR you could consider deleting the resources with the `dss.configDSS_CLI_MANAGER_LABELS` label. Perhaps in a separate PR you could also add an `ownerReference` to the `Service` so it only gets deleted IFF the `Deployment` gets deleted, as @ca-scribner suggests.

kimwnasptd: Thanks for the ping! `ownerReferences` is something I've seen in operator pattern, in which an object owns other objects (CR owns underlying objects).

I would suggest to not go with this road for now, as this is more of an indicator we need a K8s controller. And in this case it means the CLI is this controller.

tests/integration/test_dss.py
NohaIhab: agreed

src/dss/remove_notebook.py
DnPlas: I'm okay with what @kimwnasptd mentions, if we consider it for the future, sounds like it can be added in a separate PR. I wonder if you are still considering the labels, though @NohaIhab.

DnPlas (commented):
It's looking good overall, there's still an open conversation about labels. Thanks!

src/dss/remove_notebook.py
ca-scribner, line 26: What about the case where only the service exists?  In that case we try to delete the Deployment, get a 404 error, and exit without trying to delete the Service. 

I propose we refactor this to ensure we try to delete every object, even if a previous one failed.  We do this same approach [in charms](https://github.com/canonical/charmed-kubeflow-chisme/blob/8a7454fede8dbcefe65cd45f2fc04a3926b7e421/src/charmed_kubeflow_chisme/lightkube/batch/_many.py#L78).

ca-scribner: current way lgtm

misohu: +1 on this

NohaIhab: @DnPlas what's your suggestion on using labels? I think what you're suggesting would apply to #75 to remove all the resources managed by DSS, but I'm not sure how it applies here. Can you elaborate?

NohaIhab: makes sense, I've refactored accordingly

misohu (dismissed):
Good Job @NohaIhab

src/dss/remove_notebook.py
DnPlas: I was asking because you mentioned it in an earlier comment, but if that doesn't apply here, there is nothing else to discuss (:

misohu (approved):
GJ :)

PR #59, opened 2024-03-28
Update the README

kimwnasptd:
An initial run through the README so that it has some initial structure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 The Data Science Stack (DSS) aims to make it seamless for everyone to jump into an ML Environment within minutes and be able to utilise their GPUs.
 
 DSS is a ready-made environment which allows everyone to run ML workloads on the laptop.
-It gives easy access to a solution to develop and optimise ML models, that leverages the GPUs of the laptop by
+It gives easy access to a solution for developing and optimising ML models, that leverages the GPUs of the laptop by
 enabling users to utilise different ML environment images based on their needs.
 
 The DSS is a stack that includes 
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ GPU and drivers, so the containerized environments can only focus on user-space 
 
 - Containerized environment management
 - Seamless GPU utilization
-- Out-of-the box ML Environments
+- Out-of-the box ML Environments with JupyterLab
 - Easy data passing between local machine and containerized ML Environments
 - [MLFlow](https://github.com/mlflow/mlflow) for lineage tracking
 
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ The DSS is a stack that includes
 - out-of-the box containerized ML Environments
 - an intuitive CLI which streamlines the management of those containers (data-science-stack snap)
 
-The container orchestration system also handles how to integrate with the host's
+The container orchestration system also handles the integration with the host's
 GPU and drivers, so the containerized environments can only focus on user-space libraries.
 
 ## Features
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1 +1,48 @@
...
+
+## Features
+
+- Containerized environment management
+- Seamless GPU utilization
+- Out-of-the box ML Environments
+- Easy data passing between local machine and containerized ML Environments
+- [MLFlow](https://github.com/mlflow/mlflow) for lineage tracking
+
+## Requirements
```
NohaIhab, line 33: should we mention GPU drivers?

NohaIhab (commented):
I think we're missing the mention of Jupyter Notebooks

README.md
```diff
@@ -1 +1,48 @@
-# Data Science Stack
+<div align="center">
+
+# Data Science Stack ✨
+
+*Making it seamless to run GPU enabled containerized ML Environments*
+
+</div>
```
orfeas-k, line 7: We don't do that in other repos. It's fine but it's a bit out of the style we have.

README.md
```diff
@@ -1 +1,48 @@
...
+
+## Overview
+
+The Data Science Stack (DSS) aims to make it seamless for everyone to jump into an ML Environment within minutes and be able to utilise their GPUs.
```
orfeas-k, line 11:
```suggestion
The Data Science Stack (DSS) makes it seamless for everyone to jump into an ML Environment within minutes and be able to utilise their GPUs.
```
?

README.md
```diff
@@ -1 +1,48 @@
...
+
+DSS is a ready-made environment which allows everyone to run ML workloads on the laptop.
+It gives easy access to a solution to develop and optimise ML models, that leverages the GPUs of the laptop by
```
orfeas-k, line 14:
```suggestion
It gives easy access to a solution for developing and optimising ML models, that leverages the GPUs of the laptop by
```

README.md
```diff
@@ -1 +1,48 @@
...
+enabling users to utilise different ML environment images based on their needs.
+
+The DSS is a stack that includes 
+- a container orchestration system ([microK8s](https://microk8s.io/) snap)
+- out-of-the box containerized ML Environments
+- an intuitive CLI which streamlines the management of those containers (data-science-stack snap)
+
+The container orchestration system also handles how to integrate with the host's
```
orfeas-k, line 22:
```suggestion
The container orchestration system also handles the integration with the host's
```

README.md
kimwnasptd: The drivers specifically is a weird grey area. While the docs will help people on setting this up, even in a future iteration we can't handle the driver installation in the DSS (even if we had a non-strict snap).

There's also distribution issues with the drivers so the drivers (NVIDIA) can't be part of the DSS, but rather the DSS can help set those up.

I kept tho the "Seamless GPU utilisation" since this includes the work we do on the K8s side to enable the GPU

README.md
kimwnasptd: I've seen a similar README in the [oci-factory](https://github.com/canonical/oci-factory) and thus the sparkle emoji and the difference.

I'd personally like for our products (Kubeflow, DSS, MLFlow) to start have more "flashy" docs, compared to for example our charm repos.

Of course this will become a more focused effort once we also get a technical author. @orfeas-k if you don't have a hard preference against it I'd suggest we keep it like this for now and we can revisit later on once we'll be finalising the README

orfeas-k: SGTM

README.md
NohaIhab: I'm ok with this for now, as you mentioned we will have the docs on setup so that will complete the story

NohaIhab (approved):

kimwnasptd:
Thank you @NohaIhab @orfeas-k for the review!

Issue #67, opened 2024-04-08
Implement DSS `status` command


misohu:
### Why it needs to get done

The status command checks the status of key components within the DSS environment. It verifies if the MLflow deployment is ready and checks if GPU acceleration is enabled on the Kubernetes cluster by examining the labels of Kubernetes nodes for NVIDIA or Intel GPU devices.

```
dss status
```

Please use the UX [spec](https://docs.google.com/document/d/1v0wU8TBb-ui6RcjGRbLwA6XFZnWbuiVb2TPmEgwKejY/edit#heading=h.rfcm5y81sk1z) documentation .

### What needs to get done

1. Implement start command
2. Implement integration test with Microk8s

### When is the task considered done

1. Command is implemented
2. Tests are passing

PR #69, opened 2024-04-10
feat: implement dss status

misohu:
Resolves: https://github.com/canonical/data-science-stack/issues/67

The nvidia gpu status was tested only on my local machine as the runners don't have gpus.

**Steps to test locally:** 
Install and setup microk8s 
```
sudo snap install microk8s --channel=1.28/stable --classic
microk8s enable storage dns rbac gpu
microk8s config > ~/.kube/config
```

Run dss
```
export KUBECONFIG=~/.kube/config

pip install .
dss status
```

### DIFF
--- a/src/dss/status.py
+++ b/src/dss/status.py
@@ -45,6 +45,4 @@ def get_status(lightkube_client: Client) -> None:
             logger.info("GPU acceleration: Disabled")
     except Exception as e:
         logger.error(f"Failed to retrieve status: {e}")
-        logger.info("MLflow deployment: Not ready")
-        logger.info("GPU acceleration: Not detected")
         return
--- a/src/dss/status.py
+++ b/src/dss/status.py
@@ -1,2 +1,47 @@
-def get_status(lightkube_client):
-    pass
+from lightkube import Client
+
+from dss.config import DSS_NAMESPACE, MLFLOW_DEPLOYMENT_NAME
+from dss.logger import setup_logger
+from dss.utils import does_mlflow_deployment_exist, get_labels_for_node, get_service_url
+
+# Set up logger
+logger = setup_logger("logs/dss.log")
+
+
+def get_status(lightkube_client: Client):
+    """
+    Checks the status of key components within the DSS environment.
+
+    Args:
+        lightkube_client (Client): The Kubernetes client.
+    """
+    try:
+        # Check MLflow deployment
+        mlflow_ready = does_mlflow_deployment_exist(lightkube_client)
+
+        # Log MLflow deployment status and URL
+        if mlflow_ready:
+            mlflow_url = get_service_url(MLFLOW_DEPLOYMENT_NAME, DSS_NAMESPACE, lightkube_client)
+            logger.info("MLflow deployment: Ready")
+            logger.info(f"MLflow URL: {mlflow_url}")
+        else:
+            logger.info("MLflow deployment: Not ready")
+
+        # Check NVIDIA GPU acceleration
+        gpu_acceleration = False
+        node_labels = get_labels_for_node(lightkube_client)
+        if (
+            "nvidia.com/gpu.present" in node_labels
+            and "nvidia.com/gpu.deploy.container-toolkit" in node_labels
+            and "nvidia.com/gpu.deploy.device-plugin" in node_labels
+        ):
+            gpu_acceleration = True
+
+        # Log GPU status
+        if gpu_acceleration:
+            logger.info("GPU acceleration: Enabled (NVIDIA GPU)")
+        else:
+            logger.info("GPU acceleration: Disabled")
+    except Exception as e:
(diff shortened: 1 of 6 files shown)


src/dss/status.py
```diff
@@ -0,0 +1,47 @@
...
+        if (
+            "nvidia.com/gpu.present" in node_labels
+            and "nvidia.com/gpu.deploy.container-toolkit" in node_labels
+            and "nvidia.com/gpu.deploy.device-plugin" in node_labels
+        ):
+            gpu_acceleration = True
+
+        # Log GPU status
+        if gpu_acceleration:
+            logger.info("GPU acceleration: Enabled (NVIDIA GPU)")
```
orfeas-k, line 42: According to spec, I 'd expect this to output specific GPU model.

src/dss/status.py
```diff
@@ -0,0 +1,47 @@
...
+        else:
+            logger.info("GPU acceleration: Disabled")
+    except Exception as e:
+        logger.error(f"Failed to retrieve status: {e}")
+        return
```
orfeas-k, line 47: In spec I see the following which doesn't align with what is implemented.

>Output upon failure:
MLflow deployment: Not ready
GPU acceleration: Not detected

When I ran `dss status` before initialization and without gpu, it outputed:
```
2024-04-16 10:36:18 [INFO] [status] [get_status]: MLflow deployment: Not ready
2024-04-16 10:36:18 [INFO] [status] [get_status]: GPU acceleration: Disabled
```
which is not a case according to spec.

tests/integration/test_dss.py
```diff
@@ -12,6 +12,27 @@
 NOTEBOOK_NAME = "test-nb"
 
 
+def test_status_before_initialize(cleanup_after_initialize) -> None:
+    """
+    Integration test to verify 'dss status' command before initialization.
+    """
+    kubeconfig = "~/.kube/config"
```
orfeas-k, line 19: What about putting this as a global constant for the file? Are we intend to use a different kubeconfig in a test case?

tests/integration/test_dss.py
```diff
@@ -12,6 +12,27 @@
 NOTEBOOK_NAME = "test-nb"
 
 
+def test_status_before_initialize(cleanup_after_initialize) -> None:
```
orfeas-k, line 15: Do we need `cleanup_after_initialize` in every test case? At first, I concluded from that that in every test case we clean up and re-initialize but then I saw that it is not this way. If we need it, should we redesign how the clean up takes place? Should we have clean up as a separate test case (for `dss purge`) instead of a fixture?

orfeas-k (commented):
Good job, left some comments. Feel free to hmu if you 'd like to sync offline on spec details.

**EDIT**: Could you omit the kf 5525 from the PR's title since it's going to be the commit's title? Something like `feat: Implement dss status command` would provide more complete information with a quick look

src/dss/status.py
misohu: Added correct message on error ... your message is correct ... status should output exactly that when dss is not initialised and you dont have gpu .... only on failure it should say not detected.

tests/integration/test_dss.py
misohu: So lets keep the clen up as it is but then we will remove it in your purge command ?

src/dss/status.py
```diff
@@ -0,0 +1,47 @@
...
+        else:
+            logger.info("MLflow deployment: Not ready")
+
+        # Check NVIDIA GPU acceleration
+        gpu_acceleration = False
+        node_labels = get_labels_for_node(lightkube_client)
+        if (
+            "nvidia.com/gpu.present" in node_labels
+            and "nvidia.com/gpu.deploy.container-toolkit" in node_labels
+            and "nvidia.com/gpu.deploy.device-plugin" in node_labels
```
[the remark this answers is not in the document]
misohu: The first version should handle just nvidia ... later we will add intel but not in this first version

src/dss/status.py
```diff
@@ -36,12 +36,15 @@ def get_status(lightkube_client: Client):
...
 
         # Log GPU status
         if gpu_acceleration:
-            logger.info("GPU acceleration: Enabled (NVIDIA GPU)")
+            logger.info(f"GPU acceleration: Enabled ({card_name})")
         else:
             logger.info("GPU acceleration: Disabled")
     except Exception as e:
         logger.error(f"Failed to retrieve status: {e}")
+        logger.info("MLflow deployment: Not ready")
```
orfeas-k, line 48: I think there is a case where "MLflow deployment: Not ready" has already been printed from  L28 and then we reprint it here. Where is this exception expected to be raised from?

src/dss/status.py
orfeas-k: Added another comment,we can follow up the discussion there. Resolving this one.

orfeas-k (approved):

Issue #66, opened 2024-04-08
Implement DSS `list` command


misohu:
### Why it needs to get done

The list command retrieves and displays a list of Jupyter notebooks within the DSS environment. This command provides users with an overview of existing notebooks.

```
dss list
```

Example output 
```
Name           State    Image                                               URL
user-notebook  Running  kubeflownotebookswg/jupyter-scipy:v1.8.0            http://10.152.183.223/
experiment-1   Running  charmedkubeflow/jupyter-pytorch-full:1.8.0-3058193  [http://10.152.183.22](http://10.152.183.223/)5
experiment-2   Stopped  kubeflownotebookswg/jupyter-scipy:v1.8.0            (stopped)
```

Please use the UX [spec](https://docs.google.com/document/d/1v0wU8TBb-ui6RcjGRbLwA6XFZnWbuiVb2TPmEgwKejY/edit#heading=h.rfcm5y81sk1z) documentation.

### What needs to get done

1. Implement list command
2. Implement integration test with Microk8s

### When is the task considered done

1. Command is implemented
2. Tests are passing

misohu:
We are iintroducing states of the notebook lifecycle which will be tracked with `dss list` namely:
- Downloading (notebook is downloading image)
- Starting (notebook is starting so image is downloaded but the pod is not up)
- Removing (remove was executed but the notebook is still present)
- Stopping (notebook was scaled down but there is still replica)

PR #73, opened 2024-04-12
Implement `list` command

misohu:
Steps to test locally:
Install and setup microk8s
```
sudo snap install microk8s --channel=1.28/stable --classic
microk8s enable storage dns rbac gpu
microk8s config > ~/.kube/config
```

Run dss
Run dss
```
export KUBECONFIG=~/.kube/config

pip install .
dss create-notebook test-notebook --image kubeflownotebookswg/jupyter-scipy:v1.8.0 --kubeconfig $KUBECONFIG
dss list --kubeconfig $KUBECONFIG
```


ca-scribner:
I think this closes #66 but is not linked

src/dss/main.py
```diff
@@ -130,5 +134,34 @@ def status_command(kubeconfig: str) -> None:
...
+    """
+    try:
+        kubeconfig = get_default_kubeconfig(kubeconfig)
+        lightkube_client = get_lightkube_client(kubeconfig)
+        list_notebooks(lightkube_client, wide)
+    except ApiError as e:
+        logger.error(f"Error interacting with Kubernetes: {e}", err=True)
+        sys.exit(1)
+    except Exception as e:
+        logger.error(f"An error occurred: {e}", err=True)
```
ca-scribner, line 162: There's a syntax error here - `logger.error()` does not have an `err` keyword arg.  Repro this by running `dss list` without a kubeconfig

src/dss/utils.py
```diff
@@ -211,3 +228,89 @@ def get_labels_for_node(lightkube_client: Client) -> dict:
...
+    # Check for deployment deletion
+    if deletion_timestamp:
+        return DeploymentState.REMOVING
+
+    # Check pod statuses for image pulling issues and container creation
+    pods = lightkube_client.list(
+        Pod, namespace=deployment.metadata.namespace, labels=deployment.spec.selector.matchLabels
+    )
+    for pod in pods:
+        for container_status in pod.status.containerStatuses:
```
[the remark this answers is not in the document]
ca-scribner: It eventually goes away as the pod comes up, but for a little bit it has that error

src/dss/list.py
```diff
@@ -0,0 +1,54 @@
...
+                    if available_replicas
+                    else f"({DeploymentState.STOPPED})"
+                )
+            else:
+                url = f"({state.value})"
+
+            if not wide:
+                name, image, url = truncate_row(name, image, url)
+
+            notebook_list.append([name, image, url])
```
ca-scribner, line 51: iiuc the `if NOTEBOOK_LEVEL in labels:` part of this is just to filter such that we only inspect `Deployments` that include any label with the key `NOTEBOOK_LABEL`.  If yes, we can simplify this by using the label filtering on `lightkube_client.list()`:
```suggestion
    try:
        deployments = lightkube_client.list(Deployment, namespace=DSS_NAMESPACE, labels={NOTEBOOK_LABEL: lightkube.operators.exists()})
    except ApiError as e:
        logger.error(f"Failed to list notebooks: {e}")
        return

    notebook_list = []
    for deployment in deployments:
        name = deployment.metadata.name
        image = deployment.spec.template.spec.containers[0].image
        state = get_deployment_state(deployment, lightkube_client)

        # Use state to decide what to display in the URL column
        if state == DeploymentState.ACTIVE:
            available_replicas = deployment.status.availableReplicas
            url = (
                get_service_url(name, DSS_NAMESPACE, lightkube_client)
                if available_replicas
                else f"({DeploymentState.STOPPED})"
            )
        else:
            url = f"({state.value})"

        if not wide:
            name, image, url = truncate_row(name, image, url)

        notebook_list.append([name, image, url])
```

This syntax means "fetch any deployment that has any label with key NOTEBOOK_LABEL" ([ref](https://lightkube.readthedocs.io/en/latest/selectors/)).  If you prefer, we can simplify it down to `lightkube_client.list(Deployment, namespace=DSS_NAMESPACE, labels={NOTEBOOK_LABEL: None})` too, but imo the `lightkube.operators.exists()` feels more readable.

ca-scribner: Sorry, the diff viewer here makes it hard to see the changes.  Might be easier to paste this into your local file.  Its just changing the `.list()` call args, then removing the `if NOTEBOOK_LABEL in labels:` part

src/dss/utils.py
```diff
@@ -211,3 +228,89 @@ def get_labels_for_node(lightkube_client: Client) -> dict:
         raise ValueError("Expected exactly one node in the cluster")
 
     return nodes[0].metadata.labels
+
+
+def truncate_row(name: str, image: str, url: str, max_length: int = 80) -> tuple[str, str, str]:
```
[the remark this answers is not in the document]
ca-scribner: There's probably a lib somewhere that does what you want for you, but I haven't found one after a little googling...

src/dss/list.py
```diff
@@ -0,0 +1,54 @@
...
+        if NOTEBOOK_LABEL in labels:
+            name = deployment.metadata.name
+            image = deployment.spec.template.spec.containers[0].image
+            state = get_deployment_state(deployment, lightkube_client)
+
+            # Use state to decide what to display in the URL column
+            if state == DeploymentState.ACTIVE:
+                available_replicas = deployment.status.availableReplicas
+                url = (
+                    get_service_url(name, DSS_NAMESPACE, lightkube_client)
```
ca-scribner, line 41: If I have a `Notebook` deployment that does not have a `Service` and I run `dss list --kubeconfig $KUBECONFIG`, I get the following error:

```
2024-04-18 15:26:34 [ERROR] [utils] [get_service_url]: Failed to get the URL of notebook test-notebook with error code 404.
Check the debug logs for more details.
2024-04-18 15:26:34 [DEBUG] [utils] [get_service_url]: Failed to get the URL of notebook test-notebook with error: services "test-notebook" not found
2024-04-18 15:26:34 [ERROR] [main] [list_command]: An error occurred: object of type 'NoneType' has no len()
```

At minimum, `dss list` should be able to ignore this and show all other notebooks when there is one incomplete notebook present.  

Similarly, if we had a notebook `Service` without a corresponding `Deployment`, it wont show up here at all (but at least `dss list` will show all the other notebooks without issue)

Have we defined how `dss list` should behave when Notebooks are incomplete (eg: if just a notebook's deployment **or** service exist, not both)?  `dss create` has a failure message that says "Some resources might be left in the cluster. Check the status with `dss list`" which imo means `dss list` needs to show when incomplete Notebooks are deployed, too.  Or we need to change the `dss create` message

src/dss/utils.py
```diff
@@ -211,3 +228,89 @@ def get_labels_for_node(lightkube_client: Client) -> dict:
...
+                ]:
+                    return DeploymentState.DOWNLOADING
+
+    # Determine the state based on replica counts
+    if desired_replicas == 0:
+        if current_replicas == 0:
+            return DeploymentState.STOPPED
+        else:
+            return DeploymentState.STOPPING
+    else:  # desired_replicas == 1
```
ca-scribner, line 304: I think this works as is, but I'm not 100% because the assumption is being made that `desired_replicas` is always either 0 or 1.  If by some accident `desired_replicas = 2`, I'm not 100% sure what will happen here.

Rather than doing `else: # desired_replicas == 1` and assuming it can never be 2, can we handle that explicitly?  Something like:
```python
if desired_replicas == 0:
  ...
elif desired_replicas == 1:
  ...
else:
  # error!
```

that way there's no ambiguity

tests/integration/test_dss.py
```diff
@@ -108,6 +107,30 @@ def test_create_notebook(cleanup_after_initialize) -> None:
...
+        [
+            "dss",
+            "list",
+            "--kubeconfig",
+            KUBECONFIG,
+            "--wide",
+        ],
+        capture_output=True,
+        text=True,
+        timeout=60 * 4,
```
ca-scribner, line 124: will `dss list` ever take 4 minutes? :D

src/dss/utils.py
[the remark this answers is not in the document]
misohu: ok synced with ux team ... we are using librarry now with preset column widths if the lenght is above the limit the text is wrapped :)

src/dss/list.py
```diff
@@ -21,34 +24,54 @@ def list_notebooks(lightkube_client: Client, wide: bool = False) -> None:
...
+    # Create a PrettyTable object
+    table = PrettyTable()
+    table.field_names = ["Name", "Image", "URL"]
+    table.border = False
+    table.align = "l"  # Align the text to left
+
+    # Adjust column widths based on the output destination
+    if sys.stdout.isatty() and not wide:
+        # Output is to a terminal and not in wide mode
+        table._max_width = {"Name": 26, "Image": 30, "URL": 24}
```
ca-scribner, line 52: Can we increase the URL width here?  Name and Image aren't so bad when they're truncated, but URL is kinda useless.  The max width possible is 29 or 30 char, although since we set the port and it wont be https I think 26 will catch everything we need

ca-scribner: Ok I see also that if anything is wider than the column it gets wrapped rather than truncated, so the content is still available either wya.  And there is a `--wide` option.  So if you want to keep it as is that is ok, but 26 would be my preference

src/dss/list.py
```diff
@@ -21,34 +24,54 @@ def list_notebooks(lightkube_client: Client, wide: bool = False) -> None:
...
+            )
+        )
     except ApiError as e:
         logger.error(f"Failed to list notebooks: {e}")
         return
 
-    notebook_list = []
+    if not deployments:
+        # Use stdout for table so the output can be piped properly
+        print("No notebooks found in dss")
```
ca-scribner, line 40: I don't think you need "in dss" here, since you're calling `dss list`.  If you want dss in the message, I think "No dss notebooks found" is more correct

```suggestion
        print("No notebooks found")
```

src/dss/list.py
ca-scribner: k that sounds good as an interpretation.  Can we put that in the spec too?

ca-scribner (changes requested):
a few tiny things, otherwise lgtm

src/dss/list.py
misohu: Added [section](https://docs.google.com/document/d/1v0wU8TBb-ui6RcjGRbLwA6XFZnWbuiVb2TPmEgwKejY/edit#heading=h.r2mgghl5zl7j) to spec. FOr now we can keep the messages as we want to fix how the logger messages are displayed in the stdout [here](https://github.com/canonical/data-science-stack/issues/77) is the issue.

src/dss/list.py
misohu: Lets keep it for now.

src/dss/list.py
```diff
@@ -0,0 +1,77 @@
...
+        deployments = list(
+            lightkube_client.list(
+                Deployment,
+                namespace=DSS_NAMESPACE,
+                labels={NOTEBOOK_LABEL: lightkube.operators.exists()},
+            )
+        )
+    except ApiError as e:
+        logger.error(f"Failed to list notebooks: {e}")
+        return
```
orfeas-k, line 36: This will not result in exit code though, will it?

ca-scribner:
everything lgtm from the comments I had, but @orfeas-k does raise a good point about error handling

orfeas-k (approved):

ca-scribner (approved):

PR #75, opened 2024-04-16
feat: Implement `dss purge` command

### LINKED ISSUE #34: Implement DSS `purge` command

orfeas-k:
* Implement `dss purge` command
* Use `dss purge` to clean up integration tests (and thus test that it
 works as expected).
* Add unit tests for `purge` function and helper functions.

Closes #34

### Testing
Given that microk8s is already installed, run dss
```
git checkout origin/kf-5383-feat-dss-purge-command
pip install .
dss initialize --kubeconfig ~/.kube/config
dss create test-notebook
# verify resources (mlflow deployment, notebook pods and pvcs) 
dss purge --kubeconfig ~/.kube/config
# verify resources (everything has been deleted)
```

#### Verify resources in dss namespace
```bash
kubectl get ns
kubectl get po -n dss
kubectl get pvc -n dss
```

### DIFF
--- a/src/dss/main.py
+++ b/src/dss/main.py
@@ -212,6 +212,7 @@ def remove_notebook_command(name: str, kubeconfig: str):
     except RuntimeError:
         exit(1)
 
+
 @main.command(name="purge")
 @click.option(
     "--kubeconfig",
@@ -238,6 +239,9 @@ def purge_command(kubeconfig: str) -> None:
         logger.info("  dss status      to check the current status")
         logger.info("  dss logs --all  to review all logs")
         logger.info("  dss initialize  to install dss")
+        exit(1)
+    except Exception as err:
+        logger.error(f"An error occurred: {err}")
         exit(1)
 
 if __name__ == "__main__":
--- a/src/dss/main.py
+++ b/src/dss/main.py
@@ -223,25 +223,16 @@ def purge_command(kubeconfig: str) -> None:
     """
     Removes all notebooks and DSS components.
     """
-    logger.info("Executing purge command")
-
-    kubeconfig = get_default_kubeconfig(kubeconfig)
-    lightkube_client = get_lightkube_client(kubeconfig)
     try:
+        kubeconfig = get_default_kubeconfig(kubeconfig)
+        lightkube_client = get_lightkube_client(kubeconfig)
         purge(lightkube_client=lightkube_client)
-    except ApiError as err:
-        logger.error(
-            f"Failed to purge DSS components with error code {err.status.code}. Please try again."
-        )
-        logger.debug(f"Failed to purge DSS components with error {err}.")
-        logger.info("You might want to run")
-        logger.info("  dss status      to check the current status")
-        logger.info("  dss logs --all  to review all logs")
-        logger.info("  dss initialize  to install dss")
-        exit(1)
-    except Exception as err:
-        logger.error(f"An error occurred: {err}")
-        exit(1)
+    except RuntimeError:
+        click.get_current_context().exit(1)
+    except Exception as e:
+        logger.debug(f"Failed to purge DSS components: {e}.", exc_info=True)
+        logger.error(f"Failed to purge DSS components: {str(e)}.")
+        click.get_current_context().exit(1)
 
 
 if __name__ == "__main__":
--- a/src/dss/main.py
+++ b/src/dss/main.py
@@ -1,17 +1,16 @@
 import click
 from lightkube.core.exceptions import ApiError
 
-from lightkube.core.exceptions import ApiError
 from dss.config import DEFAULT_NOTEBOOK_IMAGE, RECOMMENDED_IMAGES_MESSAGE
 from dss.create_notebook import create_notebook
 from dss.initialize import initialize
 from dss.list import list_notebooks
 from dss.logger import setup_logger
 from dss.logs import get_logs
+from dss.purge import purge
 from dss.remove_notebook import remove_notebook
 from dss.status import get_status
 from dss.stop import stop_notebook
-from dss.purge import purge
 from dss.utils import KUBECONFIG_DEFAULT, get_default_kubeconfig, get_lightkube_client
 
 # Set up logger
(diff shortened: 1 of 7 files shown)


tests/integration/test_dss.py
```diff
@@ -130,26 +130,34 @@ def test_log_command(cleanup_after_initialize) -> None:
...
+    assert result.returncode == 0
+    assert (
+        "Success: All DSS components and notebooks purged successfully from the Kubernetes cluster."
+        in result.stderr
     )
 
-    # Attempt to clean up anything that initialize might create
-    # Note that .delete() does not wait on the objects to be successfully deleted, so repeating
-    # the tests quickly can still cause an issue
-    k8s_resource_handler.delete()
```
NohaIhab, line 155: it's a bit tricky testing the removal in the integration tests, removing the cleanup fixture means that if something goes wrong with the purge command then the test resources will stay there. Maybe we can keep the cleanup fixture as a fail-safe. wdyt?

tests/unit/test_utils.py
```diff
@@ -260,3 +271,74 @@ def test_does_dss_pvc_exist(lightkube_client_side_effect, context_raised, expect
...
+            namespace=namespace,
+            timeout_seconds=2,
+            interval_seconds=1,
+        )
+
+    # Assertions
+    mock_logger.info.assert_called_with(f"Waiting for namespace {namespace} to be deleted...")
+    assert mock_logger.info.call_count == logger_info_call_count
+    if error_message:
+        error_message = error_message.replace("{namespace}", namespace)
```
NohaIhab, line 334: I think you want to use [`.format`](https://docs.python.org/3.10/library/stdtypes.html#str.format) here

src/dss/purge.py
```diff
@@ -0,0 +1,55 @@
...
+    try:
+        if not does_namespace_exist(lightkube_client, DSS_NAMESPACE):
+            logger.error(
+                f"Failed to purge DSS components. Namespace `{DSS_NAMESPACE}` does not exist."
+            )
+            logger.info(
+                "You might want to run\n"
+                "  dss status      to check the current status\n"
+                "  dss logs --all  to review all logs\n"
+            )
```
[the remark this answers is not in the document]
NohaIhab: @misohu 's input can be useful here

NohaIhab (commented):
thanks @orfeas-k 
left some comments

tests/unit/test_utils.py
orfeas-k: Resolving since this part is removed.

src/dss/purge.py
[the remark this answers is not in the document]
orfeas-k: I thought about it too but I think we should not suggest to `dss initialize` because if I'm a user and I'm trying to run `purge`, it sounds a bit awkward to suggest me that I should initialize.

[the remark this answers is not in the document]
misohu: I agree with Noha maybe user accidentally run it twice and the recommendation will hint that maybe they need to reinitialize. I think we should add the dss initialize in the hint

tests/integration/test_dss.py
orfeas-k: Good point actually

src/dss/purge.py
```diff
@@ -0,0 +1,47 @@
...
+            logger.info("  dss initialize  to install dss")
+            return
+        else:
+            lightkube_client.delete(Namespace, DSS_NAMESPACE)
+            # need to wait on namespace deletion to be completed
+            wait_for_namespace_to_be_deleted(lightkube_client, DSS_NAMESPACE)
+            logger.info(
+                "Success: All DSS components and notebooks purged successfully from the Kubernetes cluster."  # noqa E501
+            )
+    except ApiError as err:
```
misohu, line 38: This funtion should be reused in dss initialize when executed with `--purge` I think we should remove this big try and catch and just do try and catch in the main with the error message.

NohaIhab: +1

orfeas-k: Pushed a commit for this. Regarding unit testing logic in `main.py` file, I gave it shot but couldn't write tests that work. As I messaged @misohu about it, I think that you can't just run click commands and that's why I was getting some weird errors (https://stackoverflow.com/questions/53203500/unittest-for-click-module). I tried using the referenced module but I 'm not sure how to mock things since mocking with MagicMock() didn't work when command run in that CliRunner. Maybe we should have an issue for adding unit tests for main, and tackle this in a separate effort.

For reference, the [errors](https://pastebin.canonical.com/p/vz853t9Ms4/) I was getting.

src/dss/main.py
```diff
@@ -116,5 +118,25 @@ def logs_command(kubeconfig: str, notebook_name: str, print_all: bool, mlflow: b
...
+    """
+    Removes all notebooks and DSS components.
+    """
+    logger.info("Executing purge command")
+
+    kubeconfig = get_default_kubeconfig(kubeconfig)
+    lightkube_client = get_lightkube_client(kubeconfig)
+    try:
+        purge(lightkube_client=lightkube_client)
+    except Exception:
```
misohu, line 138: Maybe we should also log the exception ... example [here](https://github.com/canonical/data-science-stack/pull/73/files#diff-2f57dc654b74f69ba83e508cc30ef64d0c75f5d95664138c59e693663019d467R160)

src/dss/purge.py
misohu: lets open issue for these missing tests in main

orfeas-k: https://github.com/canonical/data-science-stack/issues/82

src/dss/main.py
orfeas-k: Pushed this as well

misohu (approved):
Good job :)

misohu (approved):
looking good :)

PR #85, opened 2024-04-22
Add build and test CI for dss snap

### LINKED ISSUE #26: Create a CI for testing the DSS snap
### Why it needs to get done

We need a way of testing the DSS snap on `pull request` and `push`.

### What needs to get done

Have a Github workflow that runs testing (unit and/or integration) on the snap. The testing that has to be done is ensuring the snap builds and the CLI commands work and don't return a non-zero exit code.

Please refer to #23 for more details on what can be leveraged (actions, reusable code, etc.)

When there is a workflow that automatically runs the sanity tests described above.

ca-scribner:
Implements CI for building and testing the dss snap for push and pull_request events.  

Closes #26

### DIFF
--- a/.github/workflows/build_and_test_snap.yaml
+++ b/.github/workflows/build_and_test_snap.yaml
@@ -57,6 +57,9 @@ jobs:
         channel: '1.28/stable'
         addons: '["hostpath-storage"]'
 
+    - name: Install dependencies
+      run: python3 -m pip install tox
+
     - name: Export kubeconfig where strict snap can access it
       run: |
         mkdir -p ~/.dss
--- a/.github/workflows/build_and_test_snap.yaml
+++ b/.github/workflows/build_and_test_snap.yaml
@@ -1,0 +1,67 @@
+# Note: The trigger for GPU tests is based off the name of this workflow.
+#       If we rename this workflow, update the GPU test trigger as well.
+name: Build and test Snap
+
+on:
+  workflow_call:
+  workflow_dispatch:
+
+env:
+  SNAP_ARTIFACT_NAME: dss-snap
+  SNAP_FILE: dss.snap
+  KUBECONFIG: ~/.dss/config
+
+jobs:
+  build-snap:
+    name: Build the snap
+    runs-on: ubuntu-22.04
+    outputs:
+      snap-file: ${{ steps.snapcraft.outputs.snap }}
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - uses: snapcore/action-build@v1.2.0
+      id: snapcraft
+      with:
+        # Use a deterministic file name, so we can easily reference it later
+        snapcraft-args: "-o ${{ env.SNAP_FILE }}"
+
+    - uses: actions/upload-artifact@v4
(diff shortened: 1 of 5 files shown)


.github/workflows/build_and_test_snap.yaml
```diff
@@ -0,0 +1,67 @@
...
+        addons: '["hostpath-storage"]'
+
+    - name: Export kubeconfig where strict snap can access it
+      run: |
+        mkdir -p ~/.dss
+        cp ~/.kube/config ${{ env.KUBECONFIG }}
+
+    - name: Test dss snap
+      run: |
+        data-science-stack.dss initialize --kubeconfig ${{ env.KUBECONFIG }}
```
misohu, line 66: Lets create a tox environment for this so SolQA can just execute it on their machines. I also think we should test all the commands. I propose:
- initialize 
- run status
- create notebook
- list notebooks 
- stop notebook 
- start notebook 
- remove notebook 
- purge DSS

.github/workflows/build_and_test_snap.yaml
```diff
@@ -0,0 +1,67 @@
...
+        data-science-stack.dss create x --image kubeflownotebookswg/jupyter-scipy:v1.8.0 --kubeconfig ${{ env.KUBECONFIG }}
```
misohu, line 67: Is there a reason why we dont create the alias ? 
```
sudo snap alias  data-science-stack.dss dss
```

misohu (commented):
Good job @ca-scribner I left some comments. Please also update the description of PR.

.github/workflows/build_and_test_snap.yaml
ca-scribner: Omitted the alias before since it doesn't really change what is being tested, but now that we've moved the tests to a tox environment that directly calls `dss` as the executable I've added the alias in above in the snap building step

.github/workflows/build_and_test_snap.yaml
ca-scribner: I've added these, but I don't think we should include so much for two reasons:
* The proposed list of tests is not idempotent.  This test does not run isolated, it runs in a global microk8s.  If `dss list` fails, the test will leave the sample notebook in the cluster.  Running the test a second time will then fail because the first test didn't clean itself up (sample notebook already exists).  
*   Everything listed is tested in the `integration` environment.  What we're testing here is stuff that packaging it into a snap can break.  

imo what we need here is just:
* `dss status`

That command needs a valid kubeconfig in an accessible location, will break if there's no microk8s access, and is idempotent so we don't need to worry about it leaving fragments around when tests fail.  wdyt?

misohu: After some thinking ... your afrgument makes sense ... lets just run initialize and status.... for the gpu I will create separate python librarry integration test for deploying gpu notebook and making sure it works.

ca-scribner: @misohu and I spoke offline and are reducing this to just run `dss status` since it should fully test the snap.  But for the gpu tests added later, they'll need more.

misohu (approved):
Good job :)

PR #86, opened 2024-04-23
Initialise the documentation with readthedocs

kimwnasptd:
This PR adds a first iteration of the documentation.

Specifically:
1. It creates a `./docs` folder that holds the code from the [starter pack](https://github.com/canonical/sphinx-docs-starter-pack)
2. It adds the structure in the front page (`index.rst`), to follow the text defined in our [agreed intro](https://docs.google.com/document/d/1UqCq7pLCDUIt3HkyIwPvhbJB3u3ZMZUduur_HjWzJQI/edit)

Next steps will be to ensure the project is added and can be built with readthedocs
https://library.canonical.com/documentation/publish-on-read-the-docs#creating-a-read-the-docs-project-2

### How to test
```bash
cd docs
make run
# access localhost:8000
```

PR #91, opened 2024-04-23
docs: Update gh workflow and code to work in /docs path

kimwnasptd:
Follow up from https://github.com/canonical/data-science-stack/pull/86

The docs will need to align with the instructions in https://github.com/canonical/sphinx-docs-starter-pack#documentation-in-a-code-repository

Specifically this PR:
1. Copies the gh workflow files 
2. Updates the `.readthedocs.yaml` to use the `./docs` path
3. Adds a symlink of `.wokeignore` to root


docs/.wordlist.txt
```diff
@@ -47,3 +47,9 @@ UI
 UUID
 VM
 YAML
+DSS
+JupyterLab
+MLFlow
+GPUs
+runtime
+MicroK
```
ru-fu, line 55: You should put those in `.custom_wordlist.txt`. That way, you can update `.wordlist.txt` from the starter pack without any conflicts and keep your custom terms separate.

But that's just a convention. It will still work this way. :)

ru-fu (approved):
Looks good to me!

misohu (approved):

PR #92, opened 2024-04-23
Change snap name to `data-science-stack`

misohu:
Change the snap name to `data-science-stack`.

For local testing don't forget to create the alias.

```
snapcraft
snap install data-science-stack_<name>.snap --dangerous
sudo snap alias  data-science-stack.dss dss
dss --help
```

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,10 +1,10 @@
-name: dss
+name: data-science-stack
 summary: ML development and experimentation environment manager
 description: |
      The DSS is a tool that allows workstations users to deploy
      and manage development and experimentation environments
      in systems with GPU support.
-adopt-info: dss
+adopt-info: data-science-stack
 base: core22
 confinement: strict
 architectures:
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -16,13 +16,13 @@ plugs:
     read:
       - $HOME/.dss/config
 apps:
-  dss:
+  data-science-stack:
```
kenvandine, line 19: Actually, I think the alias request you had approved will require to app to remain "dss".  The alias for dss to data-science-stack.dss will require the app to be named dss.  I think you just need to revert this one line for it to work as expected with the alias.

kenvandine (approved):
Thanks, this looks good

ca-scribner (approved):

ca-scribner:
not 100% but I think the description needs to output the kubeconfig with `microk8s config > ~/.dss/config` so the strict snap can access it

PR #96, opened 2024-04-24
fix: modify console logs according to UX feedback

NohaIhab:
closes #77 
## Summary
Removes timestamps, module names, and function names from console logs

## Testing
```
git checkout kf-5577-refactor-logger-timestamps-function-names
pip install -e .
KUBECONFIG=~/.kube/config
dss initialize --kubeconfig $KUBECONFIG
```
Output:
```
[INFO] Executing initialize command
[INFO] Waiting for deployment mlflow in namespace dss to be ready...
[INFO] Waiting for deployment mlflow in namespace dss to be ready...
[INFO] Deployment mlflow in namespace dss is ready
[INFO] DSS initialized. To create your first notebook run the command:

dss create

Examples:
  dss create my-notebook --image=pytorch
  dss create my-notebook --image=kubeflownotebookswg/jupyter-scipy:v1.8.0
```
Open `logs/dss.log` file and you should see the file logs still have timestamps and function names

Which log levels are saved to file will be adjusted in #88

PR #97, opened 2024-04-24
fix: exclude debug logs from console logger

### LINKED ISSUE #88: Refactor the `dss` logger to have `info`, `warn` and `error` logged to console and `debug` logged to file
### Why it needs to get done

to comply with the spec and the UX team's feedback

### What needs to get done

* info warning and error logs should be print to console
* debug should be only print to file

1. console logs only have `info`, `error`, and `warn` log level
2. file logs have all log levels, including `debug`
3. tests are provided to verify the above

NohaIhab:
closes #88 

## Summary

Modifies the logger to have different log levels for console and file logs:
* console logs have `INFO` log level
* file logs have `DEBUG` log level

## Testing
```
git checkout kf-5601-refactor-logger-to-console-and-file-log-levels
pip install -e .
KUBECONFIG=~/.kube/config

#try to remove a notebook that doesn't exist
dss remove nb --kubeconfig $KUBECONFIG
```
In the console logs, you can see `INFO` and `ERROR`
In the `logs/dss.log` file, you can see `INFO`, `ERROR`, and `DEBUG`


misohu:
Lets try to create integration test which will use the logger for info error warning and debug and check that debug logs are only in the file and not in console. WDYT ?

kimwnasptd:
I would propose the same comment in this one as in https://github.com/canonical/data-science-stack/pull/96#issuecomment-2075209205, that IMO it's OK to not add these tests here as it's more testing that `logging` module is working as expected. 

@misohu wdyt?

misohu (dismissed):
After syncing with @kimwnasptd we can skip the integration tests this time :) Approving

misohu (approved):

PR #99, opened 2024-04-24
fix: send logging output to stdout

### LINKED ISSUE #87: Console logs should print to stdout instead of stderr
### Why it needs to get done

Based on feedback from the UX team, console logs intended for the users to see should be printed to stdout

### What needs to get done

refactor `logger.py` to make the console logging handler print to stdout

console logs are in stdout and integration tests verify that

NohaIhab:
closes #87 

## Summary
According to the [python docs](https://docs.python.org/3/library/logging.handlers.html#streamhandler), `StreamHandler`  sends logging output to `sys.stderr` stream by default.
This PR changes `StreamHandler` to log to `sys.stdout` stream instead.

## Testing
1. Check the output went to stderr before the change
```
git checkout main
pip install -e .
KUBECONFIG=~/.kube/config
dss status --kubeconfig $KUBECONFIG > stdout.txt 2> stderr.txt 
```
observe that stderr.txt contains the logs and stdout.txt is empty
2. cleanup std files
```
rm stderr.txt stdout.txt
```
4. Check the output now goes to stdout after the change
```
git checkout kf-5600-fix-logger-output-to-stdout
dss status --kubeconfig $KUBECONFIG > stdout.txt 2> stderr.txt 
```
observe that stdout.txt contains the logs and stderr.txt is empty


NohaIhab:
thanks @ca-scribner 
your test makes sense so I modified the PR description with instructions similar to what you did

PR #103, opened 2024-04-25
Kf 5596 gpu backed integration tests for snap

misohu:


tests/integration/test_dss.py
```diff
@@ -36,6 +41,32 @@ def test_status_before_initialize(cleanup_after_initialize) -> None:
     assert "GPU acceleration: Disabled" in result.stdout
 
 
+@pytest.mark.gpu_run
+def test_status_before_initialize_gpu(cleanup_after_initialize) -> None:
```
ca-scribner, line 45: I'm sure there's a way in pytest to have an if statuement in these gpu/cpu tests that could do:
```python
if cpu_test:
  assert "GPU acceleration: Disabled
elif gpu_test:
  ...
```

but I spent a long time and couldn't figure it out :(  We can inspect the marks using `request.node.own_markers`, but that'll show all the markers the test has, not all the markers that match the `pytest -m markers` command.  :(

ca-scribner: Actually, this does work provided we always specify a mark:

```python
@pytest.mark.parametrize(
    "is_cpu_or_gpu",
    [
        pytest.param("cpu", marks=pytest.mark.cpu_run),
        pytest.param("gpu", marks=pytest.mark.gpu_run),
    ]
)
def test_status_before_initialize(is_cpu_or_gpu, cleanup_after_initialize) -> None:
    """
    Integration test to verify 'dss status' command before initialization.
    """

    # Run the status command
    result = subprocess.run(
        ["dss", "status", "--kubeconfig", KUBECONFIG], capture_output=True, text=True
    )

    # Check if the command executed successfully
    assert result.returncode == 0

    # Check if the output indicates MLflow deployment is not ready
    assert "MLflow deployment: Not ready" in result.stdout

    if is_cpu_or_gpu == "cpu":
        # Check if the output indicates GPU acceleration is disabled
        assert "GPU acceleration: Disabled" in result.stdout
    elif is_cpu_or_gpu == "gpu":
        # Check if the output indicates GPU acceleration is enabled and captures a GPU model
        assert "GPU acceleration: Enabled" in result.stdout
        assert (
                "NVIDIA" in result.stdout
        )  # This checks for the presence of NVIDIA in the log, assuming only NVIDIA GPUs are used.
    else:
        raise ValueError(f"Invalid value for is_cpu_or_gpu: {is_cpu_or_gpu}")
```

Up to you if you want to use it or leave the tests as is.  I don't think there's a downside to using it (and using it in the other duplicated tests below)

tests/integration/test_dss.py
```diff
@@ -5,17 +5,22 @@
...
 # TODO: is there a better way to initialize this?  Maybe an optional argument to the test?
 KUBECONFIG = "~/.kube/config"
 NOTEBOOK_RESOURCES_FILE = "./tests/integration/notebook-resources.yaml"
 NOTEBOOK_NAME = "test-nb"
-NOTEBOOK_IMAGE = "kubeflownotebookswg/jupyter-scipy:v1.8.0"
 
 
+def get_pod_name_from_deployment(client, deployment_name, namespace):
+    pods = list(client.list(Pod, namespace=namespace, labels={NOTEBOOK_LABEL: deployment_name}))
+    return pods[0].metadata.name if pods else None
```
ca-scribner, line 20: There's already a helper down a the bottom - let's put this helper down there too

tests/integration/test_dss.py
```diff
@@ -108,6 +141,41 @@ def test_create_notebook(cleanup_after_initialize) -> None:
     assert deployment.status.availableReplicas == deployment.spec.replicas
 
 
+@pytest.mark.gpu_run
+def test_notebook_gpu_availability(notebook_image, cleanup_after_initialize):
```
ca-scribner, line 145: does `notebook_image` get used anywhere?  I think it might be left over from a previous iteration?

tox.ini
```diff
@@ -80,11 +80,19 @@ description = Run unit tests
...
 deps =
     -r requirements-unit.txt
 description = Run integration tests
 
+[testenv:integration-gpu]
+commands =
+    pip install .
+    pytest {[vars]tst_path}/integration/ -m gpu_run -v --tb native -s --notebook-image=kubeflownotebookswg/jupyter-pytorch-cuda-full:latest {posargs}
+deps =
+    -r requirements-unit.txt
```
ca-scribner, line 93: I missed this before.  Since this requirements is used for both unit and integration tests, should we rename to `requirements-test.txt`?

.github/workflows/setup_environment.sh
```diff
@@ -0,0 +1,17 @@
...
+    exit 1
+fi
+
+# Python setup
+apt-get update -yqq
+apt-get install -yqq python3-pip
+pip3 install tox
+
+# Microk8s setup
+snap install microk8s --channel=1.28/stable --classic
```
ca-scribner, line 15: The microk8s channel gets hidden in this script - can we expose it in the workflow?  What about having:
```suggestion
snap install microk8s --channel=$MICROK8S_CHANNEL --classic
```

and then in the workflow above call with:
```yaml
    - name: Setup operator environment
      run: sudo bash .github/workflows/setup_environment.sh
      env:
        MICROK8S_CHANNEL=1.28/stable
```

.github/workflows/setup_microk8s.sh
ca-scribner, line 1: Can we rename this?  maybe "configure_microk8s.sh"?  setup makes it sound like it should be installed here, too.  I understand why it is convenient to do this in a separate step (`newgrp` and such), but the name feels confusing

.github/workflows/setup_microk8s.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# This needs to be a separate file as newgrp will not have effect from setup_environment.sh
+microk8s enable storage dns rbac gpu
```
ca-scribner, line 4: less important than the microk8s channel above, but these could be
```suggestion
microk8s enable $MICROK8S_ADDONS
```
so they appear in the workflow like our other workflows

ca-scribner (changes requested):
generally lgtm!  a few suggestions but otherwise good

.github/workflows/aws-snap-integration-gpu.yaml
```diff
@@ -0,0 +1,73 @@
+name: GPU integration CI
+
+on:
+  schedule:
+    - cron: '0 0 * * 0'  # This will trigger the workflow at 00:00 every Sunday
+  workflow_dispatch:
```
ca-scribner, line 6: Is the plan here that we only run these on demand, not as part of the regular CI?

misohu: Yes .... we conclude on this after syncing with @kimwnasptd ... the gpu instance is quite expensive so we don't want to run it on every PR ... I understand your concern tho.

misohu:
@ca-scribner resolved your comments ... did one test gpu integration ci run [here](https://github.com/canonical/data-science-stack/actions/runs/8920339814/job/24498205785). Please check again. Thanks.

.github/workflows/aws-snap-integration-gpu.yaml
ca-scribner: After discussing offline, we've compromised a bit and now have GPU tests running on push but not pull_request.  This means when we push to main we'll at least see the CI fails. 

Longer term, we discussed using labels or something else to run GPU tests only when absolutely needed, but we haven't implemented that here

ca-scribner (dismissed):
lgtm!

ca-scribner (approved):

PR #105, opened 2024-04-25
Suggestion for GPU CI

ca-scribner:
suggested updates to #103 

This is based on #85 and extends those workflows


misohu:
This one is not longer needed as we went with separate integration tests for gpu instance

PR #113, opened 2024-04-29
Add kubeconfig caching, remove kubeconfig from indivial CLI commands

### LINKED ISSUE #37: Improve how dss manages its kubeconfig file

ca-scribner:
This adds caching of the kubeconfig between dss invocations, removing the need for users to set `--kubeconfig X` on every call.  The kubeconfig is stored in a default location that will be accessible to the dss snap.

Closes #37


ca-scribner:
FYI this is mostly reviewable, but the snap CI is failing.  I am wondering if $HOME resolves differently inside a snap vs outside?

src/dss/main.py
```diff
@@ -25,7 +25,7 @@ def main():
 @main.command(name="initialize")
 @click.option(
     "--kubeconfig",
-    help=f"Path to a Kubernetes config file. Defaults to the value of the KUBECONFIG environment variable, else to '{KUBECONFIG_DEFAULT}'.",  # noqa E501
+    help=f'Content of a Kubernetes config file defining the cluster to use.  The kubeconfig will be saved to {KUBECONFIG_DEFAULT} and overwrite any previously kubeconfig stored there.  Future `dss` commands will reuse this kubeconfig by default.  Ensure you properly quote this command, for example `dss initialize --kubeconfig="$(microk8s kubeconfig)"`.',  # noqa E501
```
misohu, line 28: Lets move the example under the whole initialize command under example section similarly as we have for `dss create`

src/dss/utils.py
```diff
@@ -87,30 +89,82 @@ def wait_for_deployment_ready(
...
+    save_location.parent.mkdir(exist_ok=True)
+
+    with open(save_location, "w") as f:
+        f.write(kubeconfig)
 
 
-def get_lightkube_client(kubeconfig: Optional[str] = None):
-    # Compute the default kubeconfig, if not specified
-    kubeconfig = KubeConfig.from_file(kubeconfig)
+def get_lightkube_client():
```
misohu, line 166: Missing docs strings and type hints.

tests/unit/test_utils.py
```diff
@@ -102,49 +113,80 @@ def __init__(self, code=400):
         super().__init__(response=_FakeResponse(code))
 
 
+@patch("dss.utils.get_kubeconfig_path")
+def test_get_kubeconfig(
```
misohu, line 117: Can we add small docstring and type hints please

tests/unit/test_utils.py
```diff
@@ -102,49 +113,80 @@ def __init__(self, code=400):
...
-        kubeconfig: path to a kubeconfig file, passed to get_lightkube_client by arg
-        kubeconfig_env_var: environment variable for kubeconfig
-        expected: expected returned value for kubeconfig
-    """
-    mock_environ_get.return_value = kubeconfig_env_var
 
-    returned = get_default_kubeconfig(kubeconfig)
-    assert returned == expected
+@patch("dss.utils.get_kubeconfig_path")
+def test_save_kubeconfig(
```
misohu, line 164: Can we add small docstring and type hints please

src/dss/main.py
misohu: also want to note that correct microk8s command is `config` so `dss initialize --kubeconfig="$(microk8s config)"`

.github/workflows/build_and_test_snap.yaml
```diff
@@ -64,9 +64,9 @@ jobs:
 
     - name: Export kubeconfig where strict snap can access it
       run: |
-        mkdir -p ~/.dss
-        cp ~/.kube/config ${{ env.KUBECONFIG }}
+        mkdir -p $(dirname ${{ env.DSS_KUBECONFIG_PATH }})
+        cp ~/.kube/config ${{ env.DSS_KUBECONFIG_PATH }}
```
misohu, line 68: I dont think we should be doing this. We expect users to use the dss cli with command 

```
dss initialize --kubeconfig="$(microk8s config)"
```

so I think we should be testing that in this CI

tests/integration/test_dss.py
```diff
@@ -8,23 +9,36 @@
...
 # TODO: is there a better way to initialize this?  Maybe an optional argument to the test?
-KUBECONFIG = "~/.kube/config"
 NOTEBOOK_RESOURCES_FILE = "./tests/integration/notebook-resources.yaml"
 NOTEBOOK_NAME = "test-nb"
 NOTEBOOK_IMAGE = "kubeflownotebookswg/jupyter-scipy:v1.8.0"
 
 
-def test_status_before_initialize(cleanup_after_initialize) -> None:
+@pytest.fixture()
+def set_dss_kubeconfig_environment_variable(monkeypatch):
```
misohu, line 21: I mentioned this also above we are never testing iinmy opinion mostly common usecase for dss initialize

```
dss initialize --kubeconfig="$(microk8s config)"
```

I think we should have also test where we are actually sending the kubeconfig as text.

misohu (changes requested):
Small comments and some request for minor test modification. Otherwise really nice job @ca-scribner

.github/workflows/build_and_test_snap.yaml
ca-scribner: Yeah I see where you're coming from on that.  The downsides I see are:

* testing with `dss initialize` means the tests change the cluster's state.  So running them a second time gives a different answer
* `dss status` is faster (doesn't need to pull mlflow, etc)
* if we put `dss initialize` in the tests, we have to pass the kubeconfig through to the tests too.  So instead of `tox -e dss-cli` people always have to call it with `tox -e dss-cli -- --kubeconfig "$(microk8s config)"`

I'm against the change, but the above points are minor so I'm not hard against it.  If you prefer, we can change the tox env and use `dss initialize` as the test instead of `dss status`

tests/integration/test_dss.py
ca-scribner: Yeah I see what you mean.  At minimum, we do need to test this method of calling `dss initialize` somewhere as we currently dont at all

src/dss/main.py
ca-scribner: How about now?

tests/integration/test_dss.py
misohu: I really like it :)

misohu (approved):
Really nice job @ca-scribner

.github/workflows/build_and_test_snap.yaml
misohu: ok We do the kubeconfig test in integration tests now so its perfect :)

misohu (changes requested):
I just tested the gpu tests and they are not correctly working ... looks like the create will always create the scipy notebook now even tho I run it with gpu integration. 

```
tox -e integration-gpu -- -vv -s
```

important log 
```
PASSED
tests/integration/test_dss.py::test_create_notebook [INFO] Executing create command
[INFO] No image is specified. Using default value kubeflownotebookswg/jupyter-scipy:v1.8.0. For more information on using a specific image, see dss create --help.
[INFO] Waiting for deployment test-nb in namespace dss to be ready...
[INFO] Waiting for deployment test-nb in namespace dss to be ready...
[INFO] Deployment test-nb in namespace dss is ready
[INFO] Success: Notebook test-nb created successfully.
[INFO] Access the notebook at http://10.152.183.160:80.

PASSED
tests/integration/test_dss.py::test_notebook_gpu_availability Executing command: kubectl exec test-nb-5cb7c6f796-4xpws --namespace dss -- python -c import torch; print('CUDA Available:', torch.cuda.is_available())
FAILED
```

as you can see create uses scipy notebook.

ca-scribner:
Issue was resolved by fixing how notebook image name is passed to tests

PR #118, opened 2024-05-02
add secrets to gpu CI

ca-scribner:
Fix the bug where we forget to pass secrets to the GPU CI during push events


misohu (approved):
Good job @ca-scribner

Issue #108, opened 2024-04-26
Documentation group task 3


misohu:
### Why it needs to get done

Write documentation for:
- (How to guide / MLflow): Access the MLFlow UI
- (How to guide / MLflow): Get the MLflow logs
- (How to guide): Access your data from DSS 
- (How to guide): List available commands of CLI
- (Reference / Explanation): Enable NVIDIA GPUs
- (Reference / Explanation): Minimum Requirements (ubuntu version, tools pre-installed, memory, cpu)
- (Reference / Explanation): DSS Artefacts (stored models, created notebook ipynb)

### What needs to get done

Write the documentation with Sphinx 

### When is the task considered done

Docs are writing and docs tests are passing.

NohaIhab:
closes by #117 and #119

PR #119, opened 2024-05-03
docs: explanation and reference

NohaIhab:
part of #108 

## Summary
* adds reference doc: Enable NVIDIA GPUs
* adds reference doc: Minimum Requirements
* adds explanation doc: DSS Artefacts


docs/reference/minimum-requirements.rst
```diff
@@ -0,0 +1,8 @@
+Minimum Requirements
+====================
+
+Before getting started with DSS, make sure your system meets the following requirements:
+
+* Ubuntu 22.04
+* Snapcraft installed
```
misohu, line 7: We dont need snapcraft to run dss snap ... the devices must have snaps but every ubuntu later than 22.04 has it

docs/reference/enable-gpu.rst
```diff
@@ -0,0 +1,32 @@
...
+          sleep 5
+        done
+
+4. Install DSS snap and initialise.
+
+    .. code-block:: bash
+ 
+        microk8s config > ~/.dss/config
+        sudo snap install data-science-stack
+        dss initialize
```
misohu, line 32: We can also add that you acn check everitime the GPU status in the cluster running `dss status`

This is the output before initialization 
```
[INFO] MLflow deployment: Not ready
[INFO] GPU acceleration: Enabled (NVIDIA-GeForce-RTX-3070-Ti)
```

this is after 

```
[INFO] MLflow deployment: Ready
[INFO] MLflow URL: http://10.152.183.105:5000
[INFO] GPU acceleration: Enabled (NVIDIA-GeForce-RTX-3070-Ti)
```

note the gpu model will be different depending on card

docs/explanation/dss-artefacts.rst
```diff
@@ -0,0 +1,10 @@
+DSS Artefacts
+=============
+
+Notebook files
+--------------
+DSS Notebook .ipynb files are stored in the DSS shared volume.
+
+Models
```
misohu, line 8: So there are also other MLflow artifacts which users can and will store e.g. experiments and runs. Should we also mention those ? Please sync with @kimwnasptd .

docs/explanation/dss-artefacts.rst
```diff
@@ -0,0 +1,10 @@
...
+------
+DSS Models are stored in MLFlow.
```
misohu, line 10: Can we add link to mlflow docs please ?  https://mlflow.org/docs/latest/models.html

misohu (changes requested):
Added small comments.

docs/reference/enable-gpu.rst
NohaIhab: good suggestion
on this note I feel this doc is becoming more of a how-to not a reference, but I'm conflicted because it's not part of the core dss
what do you guys think?

docs/reference/minimum-requirements.rst
kimwnasptd: Let's also add a note here after the list to tell users that they might need more resources, depending on the Notebook OCI Image that they would run

kimwnasptd (changes requested):
Did a pass as well and added some comments

docs/explanation/dss-artefacts.rst
kimwnasptd: And also, let's have a small note in each section how a user could move those artifacts from the DSS to their machine. We could instruct them to use the Jupyter and MLflow UIs, and for Notebooks to also access the shared volume (link to docs)

NohaIhab: >  move those artifacts from the DSS to their machine

I don't think this is accurate, the files are already on their machine
we can tell them how to access the mounted volume by linking to the access data guide, wdyt?

kimwnasptd: What if someone would need to move a versioned trained model from MLflow "outside of DSS" so that they can then upload it to CKF or anything.

We'd need a way to tell them how to get those files. It doesn't have to be something super complicated, and even telling them they can download them from UI is fine. But we need a way to tell them what are the artifacts they would be interested in and how to "get them out of DSS"

NohaIhab: I made some changes, let me know if this is sufficient

ca-scribner: I'm judging this is good enough for a first pass.  If @kimwnasptd or @misohu want more, we can add it as a separate PR

misohu (approved):

PR #129, opened 2024-05-22
docs: Use sudo in MicroK8s commands

kimwnasptd:
Right now the docs are not using `sudo` for the MicroK8s commands, neither inform users to set the groups for using the `microk8s` commands without sudo
https://microk8s.io/docs/getting-started

The above will result in errors in both
1. Initialising the MicroK8s addons (dns, gpu)
2. The DSS CLI won't be able to connect to microK8s

The PR adds `sudo` to the commands, since this is also the long term approach the K8s team will take for the new K8s cli as well.


misohu (approved):
LGTM

PR #137, opened 2024-06-14
docs: Update the tutorial to include a link to the new NVIDIA gpu how-to

afgambin:
Addresing this [Jira issue](https://warthogs.atlassian.net/browse/KF-5809).

### DIFF
--- a/docs/tutorial/getting-started.rst
+++ b/docs/tutorial/getting-started.rst
@@ -12,9 +12,9 @@ It provides easy access to a solution for developing and optimising ML models, u
 Prerequisites
 -------------
 
-* Ubuntu 22.04.
+* Ubuntu 22.04 LTS.
 * An Internet connection.
-* `Snap`_.
+* `Snap`_ installed.
 
 .. _set_microk8s:
 
@@ -27,7 +27,7 @@ All the workloads and state managed by DSS will be running on top of MicroK8s.
 
 .. note::
 
-   MicroK8s requires 540MB of memory.
+   `MicroK8s requires <https://microk8s.io/docs/getting-started>`_ 540MB of memory.
    To better accommodate workloads, a system with at least 20G of disk space and 4G of memory is recommended.
 
 You can install MicroK8s using ``snap`` as follows:
@@ -63,7 +63,7 @@ Next, you need to initialise DSS on top of MicroK8s and prepare MLflow:
 .. note::
 
    The initialisation might take a few minutes to complete.
-   While the process completes, you will see the following message:
+   While the process is in progress, you will see the following message:
    ``[INFO] Waiting for deployment my-tensorflow-notebook in namespace dss to be ready...``
    
 Launch a Notebook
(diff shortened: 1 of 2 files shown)


docs/tutorial/getting-started.rst
```diff
@@ -14,20 +14,23 @@ Prerequisites
 
 * Ubuntu 22.04.
 * An Internet connection.
-* DSS relies on `MicroK8s`_, that requires as little as 540MB of memory.
-  But to accommodate workloads, we recommend a system with at least 20G
-  of disk space and 4G of memory.
+* `Snap`_.
```
kimwnasptd, line 17: Should we have a `.` at the end of each requirement here?

docs/tutorial/getting-started.rst
```diff
@@ -14,20 +14,23 @@ Prerequisites
...
 
 DSS relies on a container orchestration system, capable of exposing the host GPUs to the workloads. 
-In this tutorial, MicroK8s is used as the orchestration system.
+`MicroK8s`_ is used as the orchestration system.
 All the workloads and state managed by DSS will be running on top of MicroK8s.
 
-You can install MicroK8s using `snap`_ as follows:
+.. note::
+
+   MicroK8s requires 540MB of memory.
```
kimwnasptd, line 30: Does microK8s have a page for minimum requirements and if yes would it make sense for us to link it here?

docs/tutorial/getting-started.rst
```diff
@@ -57,21 +61,24 @@ Next, you need to initialise DSS on top of MicroK8s and prepare MLflow:
    dss initialize --kubeconfig="$(sudo microk8s config)"
 
 .. note::
-   The `dss initialize` command might take a few minutes to complete.
+
+   The initialisation might take a few minutes to complete.
+   While the process completes, you will see the following message:
```
kimwnasptd, line 66: Nit, but when I quickly saw the word `completes` here my mind thought that this will be once the command completes and not while it's running.

I would slightly change this to `Whilte the process is in progress, you'll...`

kimwnasptd (commented):
Looks pretty good @afgambin!

docs/tutorial/getting-started.rst
afgambin: Changed!

docs/tutorial/getting-started.rst
afgambin: Linking is always helpful. Added.

docs/tutorial/getting-started.rst
afgambin: We don't have an agreement yet around this in the style guide. My usual (and pretty standard) approach is we use period if it is a sentence. No period if it's just one word/link. In the case of a list of bullet points, if there is at least one sentence per list (it uses period), then all bullet points goes with period at the end.

kimwnasptd: ACK! As long as we follow it everywhere

kimwnasptd (approved):

Issue #152, opened 2024-07-24
Docs don't clearly expose the minimum disk requirements


kimwnasptd:
### Bug Description

Right now the docs mention that minimum disk size, because of MicroK8s is 20Gb. But the default notebooks of DSS are ~10Gb each. So users will need more space to be able to run them

https://documentation.ubuntu.com/data-science-stack/en/latest/tutorial/getting-started/
![image](https://github.com/user-attachments/assets/2562e070-4661-4793-8f77-06604e761eb4)

### To Reproduce

Try to follow follow the guide and create a VM with 20Gb

If you then try to create 2 notebooks you'll hit the issue posted in https://github.com/canonical/data-science-stack/issues/151

### Environment

A VM with 20Gb

### Relevant Log Output

```shell
<none>
```

kimwnasptd:
@afgambin I'd propose that we have ~50Gb as minimum requirement.

How do you think we should re-organise the page for this?

afgambin:
I'll do a quick fix to that note box next week

afgambin:
Closing this as it was already addressed here: https://github.com/canonical/data-science-stack/pull/154

PR #154, opened 2024-07-29
docs: tweaks to tutorial based on user feedback

afgambin:
Addressing user feedback and this [issue](https://github.com/canonical/data-science-stack/issues/152).

### DIFF
--- a/docs/tutorial/getting-started.rst
+++ b/docs/tutorial/getting-started.rst
@@ -17,9 +17,6 @@ Prerequisites
 * `Snap`_ installed.
 * 50GB of disk space is recommended. This includes the `requirements <https://microk8s.io/docs/getting-started>`_ for MicroK8s.
 
-.. note::
-   All the code commands provided throughout this guide can be run from any host terminal.
-
 .. _set_microk8s:
 
 Setting up MicroK8s


docs/tutorial/getting-started.rst
```diff
@@ -15,6 +15,10 @@ Prerequisites
 * Ubuntu 22.04 LTS.
 * An Internet connection.
 * `Snap`_ installed.
+* 50GB of disk space is recommended. This includes the `requirements <https://microk8s.io/docs/getting-started>`_ for MicroK8s.
+
+.. note::
+   All the code commands provided throughout this guide can be run from any host terminal.
```
afgambin, line 21: Is this correct?

docs/tutorial/getting-started.rst
```diff
@@ -15,6 +15,10 @@ Prerequisites
 * Ubuntu 22.04 LTS.
 * An Internet connection.
 * `Snap`_ installed.
+* 50GB of disk space is recommended. This includes the `requirements <https://microk8s.io/docs/getting-started>`_ for MicroK8s.
```
afgambin, line 18: I added this info here, as part of the prerequisites.

docs/tutorial/getting-started.rst
```diff
@@ -64,7 +63,13 @@ Next, you need to initialise DSS on top of MicroK8s and prepare MLflow:
 
    The initialisation might take a few minutes to complete.
    While the process is in progress, you will see the following message:
-   ``[INFO] Waiting for deployment my-tensorflow-notebook in namespace dss to be ready...``
+   ``[INFO] Waiting for deployment mlflow in namespace dss to be ready...``
```
afgambin, line 66: Got this feedback from a user, before it was wrong. First time you initialize, you see `mlflow`. Once you create a notebook, then `Waiting for deployment <your_notebook>` appears.

docs/tutorial/getting-started.rst
kimwnasptd: Looks good!

docs/tutorial/getting-started.rst
kimwnasptd: The note feels a bit weird to me when I read it initially. What do we want to convey?

docs/tutorial/getting-started.rst
kimwnasptd: Ah good point, it should be mlflow initially

docs/tutorial/getting-started.rst
afgambin: @kimwnasptd  The place where the user is expected to run the tutorial from. Is that info trivial/assumed? If so, we can remove it.

Thinking, for instance, a non-linux data scientist, coming from Windows, not used to run stuff from a CLI but using UIs.

kimwnasptd: I think in this case it's trivial and would suggest to drop :)

afgambin:
@kimwnasptd it should be good to go now. If you can pls approve and merge.

kimwnasptd (approved):

kimwnasptd:
Thanks for picking this up @afgambin!

PR #157, opened 2024-08-01
ci(upload-charm): use charmcraft from latest/candidate

orfeas-k:
Use charmcraft from latest/candidate in upload-charm in order to unify the channel used during integration tests and publishing.
  Ref canonical/bundle-kubeflow#993


orfeas-k:
Closing since this was opened by accident by the canon run.

Issue #150, opened 2024-07-22
Write `dss` doc on how to run simple calculations with Intel ML frameworks in DSS


misohu:
### Why it needs to get done

This document should describe how to run some ML workloads inside IPEX and ITEX notebooks. 

### What needs to get done

- Write a documentation about how to dpeloy Intel images on DSS.

### When is the task considered done

- Documentation is written and tested.

PR #164, opened 2024-08-20
Intel computation guide

misohu:
Closes: https://github.com/canonical/data-science-stack/issues/150

Add docs with examples for ITEX and IPEX usage.


afgambin:
@misohu I am not really sure this needs to be a dedicated guide. Can we double check with Andreea (can't tag her here) and @kimwnasptd  if we are missing something here? If all we want is to document some refs for the user to run Intel-based ML examples, then I would add a note box in the [Create an Intel GPU-enable notebook section](https://documentation.ubuntu.com/data-science-stack/en/latest/how-to/jupyter-notebook/#create-an-intel-gpu-enabled-notebook) we just published. At the bottom of the section, add a note box saying something like:

"Once created, you can access it (with hyperlink to the section) and run ML workloads on Intel hardware with DSS. See IPEX (with hyperlink) and ITEX (hyperlink) for detailed examples using the Intel extensions for Pytorch and Tensorflow respectively."

afgambin (approved):
Did some final tweaks here and there. Lgtm now @misohu !

Issue #159, opened 2024-08-07
Remove repeated output while waiting for initialization. 


misohu:
### Bug Description

Every time we initialize DSS it outputs 

```
[INFO] Waiting for deployment mlflow in namespace dss to be ready...
```

We want to output it just once.

### To Reproduce

sudo snap install dss
dss initialize --kubeconfig=$(microk8s config)

### Environment

any

### Relevant Log Output

```shell
[INFO] Executing initialize command
[INFO] Storing provided kubeconfig to /home/pocik/snap/data-science-stack/16/.dss/config
[INFO] Waiting for deployment mlflow in namespace dss to be ready...
[INFO] Waiting for deployment mlflow in namespace dss to be ready...
[INFO] Waiting for deployment mlflow in namespace dss to be ready...
[INFO] Waiting for deployment mlflow in namespace dss to be ready...
[INFO] Deployment mlflow in namespace dss is ready
[INFO] DSS initialized. To create your first notebook run the command:
...
```

PR #170, opened 2024-08-28
Remove repeated output while waiting for deployment

misohu:
Closes: https://github.com/canonical/data-science-stack/issues/159

Because there is shared function for waiting for any deployment to be ready we only had to change code in one place :) 

Manual test:
```
git clone <repo>
cd <repo>
pip install .
dss initialize --kubeconfig="$(sudo microk8s config)"
dss create my-notebook --image=pytorch
```

The message `Waiting for deployment <deployment-name> in namespace dss to be ready...` should appear only once in both cases.

### DIFF
--- a/src/dss/utils.py
+++ b/src/dss/utils.py
@@ -99,8 +99,11 @@ def wait_for_deployment_ready(
             raise TimeoutError(
                 f"Timeout waiting for deployment {deployment_name} in namespace {namespace} to be ready"  # noqa E501
             )
-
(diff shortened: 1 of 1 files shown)


rgildein (commented):
I think it is useful to see that something is in progress, can we do it as debug log? WDYT?

misohu:
Thanks @rgildein that was a good idea ... I have swapped it for debug ... debug does not go to console so its perfect.

mvlassis (approved):

PR #171, opened 2024-09-04
docs: tweaks to navigation headings

afgambin:
Some tweaks to update the way we present the navigation headings to follow the Diátaxis practice.


afgambin:
@misohu if you can have a look :)

misohu (approved):
Went through the docs :) no problems changes make sense.

PR #185, opened 2024-10-28
Build for arch=all to run on all architectures

kenvandine:
There is nothing architecture specific, so we should build to run on all.


deusebio:
As discussed offline, we currently only provide out-of-the-box support for AMD. Thank you @kenvandine  for raising this, nevertheless!

PR #190, opened 2024-12-02
Remove assertion that GPU acceleration is disabled in CPU tests

### LINKED ISSUE #188: Integration CPU tests should not assert that GPU acceleration is disabled
The CPU integration tests should not depend on whether or not the underlying k8s has been setup to use GPU acceleration.  It is reasonable for a tester to have gone through all the documentation and followed the instructions to enable GPU support, and then try to run a quick test for the CPU before trying to run the suite for the GPU.  It is unreasonable to expect that the tester will revert the enabling of GPU to run the CPU tests.

The following assertions should be removed:

https://github.com/canonical/data-science-stack/blob/2a10edcfc5280e8d481a8c8ddf645c80684cde37/tests/integration/test_dss.py#L64

https://github.com/canonical/data-science-stack/blob/2a10edcfc5280e8d481a8c8ddf645c80684cde37/tests/integration/test_dss.py#L230

motjuste:
Fixes #188

### DIFF
--- a/tests/integration/test_dss.py
+++ b/tests/integration/test_dss.py
@@ -60,8 +60,8 @@ def test_status_before_initialize(
     assert "MLflow deployment: Not ready" in result.stdout
 
     if is_cpu_or_gpu == "cpu":
-        # Check if the output indicates GPU acceleration is disabled
-        assert "GPU acceleration: Disabled" in result.stdout
+        # we don't care if gpu acceleration is enabled or not
+        pass
     elif is_cpu_or_gpu == "gpu":
         # Check if the output indicates GPU acceleration is enabled and captures a GPU model
         assert "GPU acceleration: Enabled" in result.stdout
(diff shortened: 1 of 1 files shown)


motjuste:
I don't think the failing checks are due to my changes ... one is due to some timeout, and the other due to some wrong version for the github action.  I don't have permission to re-trigger them.

tests/integration/test_dss.py
misohu, line 1: Instead of leaving empty `if` body I propose to remove the `if is_cpu_or_gpu == "cpu":` completely. So it will be just 

```
if is_cpu_or_gpu == "gpu":
    # Check if the output indicates GPU acceleration is enabled and captures a GPU model
    assert "GPU acceleration: Enabled" in result.stdout
```

This applies for both if statements.

misohu (approved):
Good job thx :)

Issue #213, opened 2025-03-10
DSS snap tries to create a `logs` directory in the current working directory


motjuste:
### Bug Description

AFAICT, DSS is a strict snap.  But, upon running any `dss` command, it tries to create a `logs` directory in the current-working-directory from where the command is run.  This is not only not expected, it can cause failures if `dss` does not have access / permissions to write to the current-working-directory.

### To Reproduce

1. `sudo snap install data-science-stack --channel 1.0/edge`
2. `sudo mkdir /tmp/somewhere`
3. `cd /tmp/somewhere`
4. `dss` - Fails

### Environment

- Ubuntu Jammy
  `Linux nvidia-dgx-station-c25989 5.15.0-134-generic #145-Ubuntu SMP Wed Feb 12 20:08:39 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux`

### Relevant Log Output

```shell
$ dss
Traceback (most recent call last):
  File "/snap/data-science-stack/36/bin/dss", line 5, in <module>
    from dss.main import main
  File "/snap/data-science-stack/36/lib/python3.10/site-packages/dss/main.py", line 4, in <module>
    from dss.create_notebook import create_notebook
  File "/snap/data-science-stack/36/lib/python3.10/site-packages/dss/create_notebook.py", line 19, in <module>
    from dss.remove_notebook import remove_notebook
  File "/snap/data-science-stack/36/lib/python3.10/site-packages/dss/remove_notebook.py", line 8, in <module>
    from dss.utils import does_notebook_exist
  File "/snap/data-science-stack/36/lib/python3.10/site-packages/dss/utils.py", line 21, in <module>
    logger = setup_logger("logs/dss.log")
  File "/snap/data-science-stack/36/lib/python3.10/site-packages/dss/logger.py", line 52, in setup_logger
    os.makedirs(os.path.dirname(log_file_path))
  File "/usr/lib/python3.10/os.py", line 225, in makedirs
    mkdir(name, mode)
PermissionError: [Errno 13] Permission denied: 'logs'
```

PR #218, opened 2025-04-03
fix: store dss logs in snap common folder

misohu:
Closes: https://github.com/canonical/data-science-stack/issues/213

Stores the logs in the `/var/snap/data-science-stack/common/logs` folder

To test this 
- remove the old version of the snap `sudo snap remove data-science-stack`
- build from this branch `snapcraft`
- deploy the snap `sudo snap install data-science-stack_0.1-<hash>_amd64.snap --dangerous`
- run some dss command e.g. `data-science-stack.dss status` (it should complain about config)
- check that the logs file is created at `/var/snap/data-science-stack/common/logs/dss.log`


src/dss/logger.py
```diff
@@ -17,52 +17,51 @@ def format(self, record: logging.LogRecord) -> str:  # noqa: A003
...
 
 def setup_logger(
-    log_file_path: str, file_log_level: int = logging.DEBUG, console_log_level: int = logging.INFO
+    log_file_path: str = None,
+    file_log_level: int = logging.DEBUG,
+    console_log_level: int = logging.INFO,
 ) -> logging.Logger:
     """
-    Set up a logger with both file and console handlers.
+    Set up a logger with optional file and console handlers.
```
orfeas-k, line 25: handlers? Or arguments?

src/dss/logger.py
```diff
@@ -17,52 +17,51 @@ def format(self, record: logging.LogRecord) -> str:  # noqa: A003
...
 
     Args:
-        log_file_path (str): Path to the log file.
-        file_log_level (int, optional): Logging level for file logs. Defaults to logging.DEBUG.
-        console_log_level (int, optional): Logging level for console logs. Defaults to logging.INFO.
+        log_file_path (str): Path to the log file. If None, uses $SNAP_COMMON/logs/dss.log.
+        file_log_level (int, optional): Logging level for file logs.
+        console_log_level (int, optional): Logging level for console logs.
```
orfeas-k, line 30: Shouldn't default values be part of the doc string?

src/dss/logger.py
```diff
@@ -17,52 +17,51 @@ def format(self, record: logging.LogRecord) -> str:  # noqa: A003
...
         )
+        console_handler = logging.StreamHandler(sys.stdout)
+        console_handler.setLevel(console_log_level)
+        console_handler.setFormatter(console_formatter)
+        logger.addHandler(console_handler)
 
-        # Formatter for file - includes [INFO]
+        # Setup file handler if path is provided or fallback to SNAP_COMMON
+        if log_file_path is None:
+            snap_common = os.environ.get("SNAP_COMMON", "/tmp")
```
orfeas-k, line 52: Is there a chance that this targets `/tmp` but the [install hook](https://github.com/canonical/data-science-stack/pull/218/files#diff-e58b9a041906b37e03356984486699655b2d559fa8dc4d95b340a3b5be7a9a42R7) targets a different place (ie snap_commonis not set when the code is executed, but it is during the install hook)? Do you think this could pose a problem for the execution?

orfeas-k (commented):
Small comments, mostly LGTM. Tested and the issue seems to have been fixed!

src/dss/logger.py
misohu: This is not a problem as the python in dss runs as user so it should have access even without the hook.

orfeas-k (approved):

